PR #12, opened 2015-04-01
new kafka consumer

jettify:
Basic port for new kafka consumer for review purposes.


Drizzt1991:
it is no longer compatible with current, rewritten codebase.

PR #15, opened 2016-02-16
Refactor Producer

Drizzt1991:


aiokafka/conn.py
```diff
@@ -17,21 +24,36 @@ def create_conn(host, port, *, loop=None):
...
 
+    @asyncio.coroutine
     def _connect(self):
-        self._reader, self._writer = yield from asyncio.open_connection(
-            self.host, self.port, loop=self._loop)
-        self._read_task = asyncio.async(self._read(), loop=self._loop)
+        future = asyncio.open_connection(self.host, self.port, loop=self._loop)
+        self._reader, self._writer = yield from asyncio.wait_for(
+            future, self._config['request_timeout_ms']/1000)
+        self._read_task = asyncio.ensure_future(self._read(), loop=self._loop)
```
jettify: Not backward compatible, not all version of python 3.4.x support `ensure_future`

aiokafka/conn.py
```diff
@@ -41,14 +63,40 @@ def host(self):
...
+    def send(self, request, expect_response=True):
+        correlation_id = self._next_correlation_id()
+        header = RequestHeader(request,
+                               correlation_id=correlation_id,
+                               client_id=self._config['client_id'])
+        message = b''.join([header.encode(), request.encode()])
+        size = self.HEADER.pack(len(message))
+        try:
+            self._writer.write(size)
+            self._writer.write(message)
```
jettify: any reason to make two writes instead of one? we can save one system calls here

aiokafka/conn.py
```diff
@@ -41,14 +63,40 @@ def host(self):
...
+
         fut = asyncio.Future(loop=self._loop)
-        if no_ack:
+        if not expect_response:
             fut.set_result(None)
             return fut
-        self._requests.append(fut)
-        return fut
+        self._requests.append((correlation_id, request.RESPONSE_TYPE, fut))
+        return asyncio.wait_for(fut, self._config['request_timeout_ms']/1000)
```
jettify: you should pass loop explicilty

jettify:
Can we split this PR in smaller ones? It is hard to review...

aiokafka/client.py
```diff
@@ -1,489 +1,274 @@
...
-
-            except Exception as e:
-                log.warning("Could not send request [%r] to server %s:%i, "
-                            "trying next server: %s",
-                            request_id, host, port, e)
-
-        raise KafkaUnavailableError("All servers failed to process request")
+    def close(self):
+        if self._sync_task:
+            self._sync_task.cancel()
```
jettify: Proper task cancel is required, you need to `yield from self._sync_task` somewhere

aiokafka/conn.py
asvetlov: We can add something like https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/helpers.py#L17-L20

aiokafka/conn.py
asvetlov: Good point! In aiohttp I've joined calls like this into one already.

aiokafka/conn.py
asvetlov: +1

asvetlov:
Honestly I did not dig deep inside proposed changes -- the patch is huge.
Splitting PR into small pieces is desirable but I can live with big change if splitting is too hard for you, guys.

Drizzt1991:
Sorry, I'm kinda used to making "pre-review" pull-requests into separate branches (it's just so easy to read it on GitHub), this should not have been reviewed yet, we're still prototyping. Sorry

PR #16, opened 2016-02-25
Dockerized tests

dmrz:
Isolated zookeeper and kafka services within a docker container with multiversion builds suport.


aiokafka/__init__.py
```diff
@@ -1,7 +1,5 @@
 __version__ = '0.0.1'
 
-from .client import AIOKafkaClient
-from .producer import SimpleAIOProducer, KeyedAIOProducer
-from .consumer import SimpleAIOConsumer
-
-(AIOKafkaClient, SimpleAIOProducer, KeyedAIOProducer, SimpleAIOConsumer)
+from .client import AIOKafkaClient  # noqa
```
asvetlov: Please revert the change -- it should be done in different PR if any.
Honestly I doubt we need it at all.

asvetlov:
I support the intention but don't like implementation.
I believe pytest fixtures better fit for working with kafka/zookeeper

TarasLevelUp:
Hi, it had a custom fixtures, based on nose. It didn't use pytest.
I do think, that the implementation has issues, but mind asking why do you think, that fixture suits better for kafka/zookeeper?

asvetlov:
I think so. It requires converting tests into pytest style but I believe it worth to be done.
We have only 21 test for now.

See http://asvetlov.github.io/pytest-slides/#/9 for describing what I mean by pytest fixtures for docker

jettify:
big plus for the fixtures

Drizzt1991:
Ok, now I kinda get where you are getting.
@asvetlov You're ok with running it inside a docker container, but would prefere it to be set up using pytest fixtures instead of a bash script?

Drizzt1991:
And I didn't kinda understand the session scope of docker's life. When does a session drop and reload in pytest?

jettify:
That depends on your scope settings, fixture is created on first request and persists depends on scope. For instance we manually execute test that does not depends on docker fixture, we do not wait on docker  star.

asvetlov:
Yes, I suggest dockerized pytests.
Session scope is for the whole test suite run.
Pushing docker fixtures into session scope allows to don't recreate containers for every test -- it's usually a relative slow procedure.

Drizzt1991:
There is no good Kafka 9.0 container in the HUB. How about we create an aiolibs or something account on Docker and have our custom containers pushed there?

asvetlov:
No problems. Could you prepare Dockerfile, I'll register it on the weekend.

Drizzt1991:
I will consult Dmitry about it. We based current solution on "spotify/kafka" cause it contains both Zookeeper and Kafka in 1 container, I think it's ok for tests.

Drizzt1991:
Will open another PR for this.

PR #17, opened 2016-02-25
Rewritten producer, client and connection.

Drizzt1991:


Drizzt1991:
Consumer is not part of this PR and is removed from coverage/tests. We will add consumer as 2 separate PR later - coordinator and consumer/fetcher

aiokafka/client.py
```diff
@@ -1,489 +1,348 @@
...
+            break
+        else:
+            log.error('Unable to update metadata from %s', nodeids)
+            self.cluster.failed_update(None)
+            return False
+        return True
 
     @asyncio.coroutine
-    def send_metadata_request(self, payloads=()):
+    def fetch_all_metadata(self):
```
jettify: this function is almost copy paste of `force_metadata_update`, I think could could be reused.

aiokafka/client.py
```diff
@@ -1,489 +1,348 @@
...
+
+        # kafka kills the connection when it doesnt recognize an API request
+        # so we can send a test request and then follow immediately with a
+        # vanilla MetadataRequest. If the server did not recognize the first
+        # request, both will be failed with a ConnectionError that wraps
+        # socket.error (32, 54, or 104)
+        conn = yield from self._get_conn(node_id)
+        for version, request in test_cases:
+            try:
+                if not conn.connected():
```
jettify: `conn` could be `None`

aiokafka/client.py
```diff
@@ -1,489 +1,348 @@
...
 
-        Return
-        ======
-        list of ProduceResponse, in the order of input payloads
+        Arguments:
+            topics (list of str): topics to track
         """
+        if set(topics).difference(self._topics):
+            self._topics = set(topics)
+            return ensure_future(
```
jettify: What logic behind this method? It returns awaitable or `None` so in one cases I can  use it with `yield from` in others not?

fabregas: `set_topics` method will be used in sync callback from `kafka.cluster.ClusterMetadata`, so we running metadata update asynchronously.
I removed return statement from this method, bcs no one method would expect the completion of the task.

aiokafka/producer.py
```diff
@@ -1,162 +1,318 @@
...
+
+    def __init__(self, *, loop, bootstrap_servers='localhost',
+                 client_id=None,
+                 metadata_max_age_ms=300000, request_timeout_ms=40000,
+                 api_version='auto', acks=1,
+                 key_serializer=None, value_serializer=None,
+                 compression_type=None, batch_size=16384,
+                 partitioner=DefaultPartitioner(), max_request_size=1048576):
+
+        assert acks in (0, 1, -1, 'all'), "Invalid ACKS parameter"
```
Drizzt1991: Change those from asserts to normal `ValueError`'s. It is a public API.

Drizzt1991: Add a check for version. It's better for consutuctor to raise a `ValueError`, than start raising an AssertionError

aiokafka/producer.py
```diff
@@ -1,162 +1,318 @@
...
+            client_id=client_id, metadata_max_age_ms=metadata_max_age_ms,
+            request_timeout_ms=request_timeout_ms)
+        self._metadata = self.client.cluster
+
+        self._closed = False
 
     @asyncio.coroutine
-    def _next_partition(self, topic):
+    def start(self):
+        log.debug("Starting the Kafka producer")  # trace
```
Drizzt1991: Why is this debug call before docstring

Drizzt1991:
Minor fixes and LGTM. Any more options?

aiokafka/client.py
```diff
@@ -1,489 +1,335 @@
...
+                strings) that the consumer should contact to bootstrap initial
+                cluster metadata. This does not have to be the full node list.
+                It just needs to have at least one broker that will respond to
+                Metadata API Request. Default port is 9092. If no servers are
+                specified, will default to localhost:9092.
+            client_id (str): a name for this client. This string is passed in
+                each request to servers and can be used to identify specific
+                server-side log entries that correspond to this client. Also
+                submitted to GroupCoordinator for logging with respect to
+                consumer group administration. Default: 'kafka-python-{ver}'
```
kxepal: Default client_id actually is different, isn't it?

aiokafka/producer.py
```diff
@@ -1,162 +1,318 @@
...
+        bootstrap_servers: 'host[:port]' string (or list of 'host[:port]'
+            strings) that the producer should contact to bootstrap initial
+            cluster metadata. This does not have to be the full node list.
+            It just needs to have at least one broker that will respond to a
+            Metadata API Request. Default port is 9092. If no servers are
+            specified, will default to localhost:9092.
+        client_id (str): a name for this client. This string is passed in
+            each request to servers and can be used to identify specific
+            server-side log entries that correspond to this client.
+            Default: 'kafka-python-producer-#' (appended with a unique number
```
kxepal: Here also default client_id doesn't match the actual one.

PR #18, opened 2016-03-01
Group coordinator for aiokafka Consumer

fabregas:
AIOConsumerCoordinator class implements group management for a single group member by interacting with a designated Kafka broker (the coordinator).


popravich:
Does any of PRs conforms PEP8? )
I see you install `flake8` on travis but you don't use it

popravich:
Its a question to @Drizzt1991

Drizzt1991:
We didn't change any behaviour there. `make cov` should still do it...

Drizzt1991:
Hmm, don't see it there, lol. Maybe a bug in previous code. @fabregas change the Makefile to run flake on `make cov` as in `make test`

aiokafka/coordinator/base.py
```diff
@@ -0,0 +1,525 @@
...
+                return resp
+            else:
+                raise error_type()
+
+    @asyncio.coroutine
+    def coordinator_unknown(self):
+        """Check if we know who the coordinator is
+        and have an active connection
+
+        Side-effect: reset coordinator_id to None if connection failed
```
Drizzt1991: What does this mean?

Drizzt1991:
I feel that ``python-kafka`` is too much ``Java`` style... Rly, do we even need so much abstraction.

Drizzt1991:
What I don't like is the `SubscriptionState` class. It uses callback's for management, and it's hard to understand all of the flow here.

Drizzt1991:
Having a BaseCoordinator is strange too, we only have 1 implementation - GroupCoordinator. And it's private API, we can add base class if we want another implementation in the future.

aiokafka/coordinator/base.py
```diff
@@ -0,0 +1,523 @@
...
+                log.info(
+                    "Heartbeat failed: local member_id was not recognized;"
+                    " resetting and re-joining group")
+                self.member_id = JoinGroupRequest.UNKNOWN_MEMBER_ID
+                self.rejoin_needed = True
+            except Errors.KafkaError as err:
+                log.error("Heartbeat failed: %s", err)
+            else:
+                log.debug("Received successful heartbeat response.")
+                last_ok_heartbeat = time.time()
```
Drizzt1991: It's better to set initial time of request

aiokafka/coordinator/base.py
```diff
@@ -0,0 +1,523 @@
...
+
+            try:
+                yield from self.ensure_active_group()
+            except Errors.KafkaError as err:
+                log.debug("Skipping heartbeat: no active group: %s", err)
+                last_ok_heartbeat = time.time()
+                sleep_time = retry_backoff_time
+                continue
+            else:
+                sleep_time = hb_interval
```
Drizzt1991: Ghost? Will always be rewritten below...

aiokafka/coordinator/base.py
```diff
@@ -0,0 +1,523 @@
...
+        retry_backoff_time = self._retry_backoff_ms / 1000
+        sleep_time = hb_interval
+
+        while True:
+            yield from asyncio.sleep(sleep_time, loop=self.loop)
+
+            try:
+                yield from self.ensure_active_group()
+            except Errors.KafkaError as err:
+                log.debug("Skipping heartbeat: no active group: %s", err)
```
Drizzt1991: This is quite a big error. Set level to ERROR i think

aiokafka/coordinator/base.py
```diff
@@ -0,0 +1,523 @@
...
+            else:
+                log.info("LeaveGroup request succeeded")
+
+        self.generation = OffsetCommitRequest.DEFAULT_GENERATION_ID
+        self.member_id = JoinGroupRequest.UNKNOWN_MEMBER_ID
+        self.rejoin_needed = True
+
+    @asyncio.coroutine
+    def _heartbeat_task_routine(self):
+        last_ok_heartbeat = time.time()
```
asvetlov: Use `loop.time()` instead.

aiokafka/__init__.py
```diff
@@ -8,7 +8,7 @@
 
 from .client import AIOKafkaClient  # noqa
 from .producer import AIOKafkaProducer  # noqa
-
 # from .consumer import SimpleAIOConsumer
```
Drizzt1991: Remove this import. I don't think it's needed

Drizzt1991:
Ok, now I like it more. It's cleaner.

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+from kafka.protocol.group import (
+    HeartbeatRequest, JoinGroupRequest,
+    LeaveGroupRequest, SyncGroupRequest)
+
+from aiokafka import ensure_future
+
+log = logging.getLogger(__name__)
+
+
+class AIOConsumerCoordinator(object):
```
Drizzt1991: Maybe rename it to GroupCoordinator.

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+    """
+    AIOConsumerCoordinator implements group management for single group member
+    by interacting with a designated Kafka broker (the coordinator). Group
+    semantics are provided by extending this class.  See ConsumerCoordinator
```
Drizzt1991: remove this `See ConsumerCoordinator` comment

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+    for example usage.
+
+    From a high level, Kafka's group management protocol consists of the
+    following sequence of actions:
+
+    1. Group Registration: Group members register with the coordinator
+       providing their own metadata
+       (such as the set of topics they are interested in).
+
+    2. Group/Leader Selection: The coordinator select the members of the group
```
Drizzt1991: Change to `The coordinator (one of kafka nodes)`

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+       and chooses one member as the leader.
```
Drizzt1991: `chooses one member (one of client's)`

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+
+    3. State Assignment: The leader collects the metadata from all the members
```
Drizzt1991: I would presume it collects this data by a connection to those clients. Lets rephrase it to `The leader receives metadata for all members and assigns partitions to them`

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+       of the group and assigns state.
+
+    4. Group Stabilization: Each member receives the state assigned by the
+       leader and begins processing.
+    """
```
Drizzt1991: We could mention something like: between each phase coordinator awaits all clients to respond. If some do not respond in time - it will revoke their membership

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+                 ):
+        """Initialize the coordination manager.
+
+        Parameters:
+            client (AIOKafkaClient): kafka client
+            subscription (SubscriptionState): instance of SubscriptionState
+                located in kafka.consumer.subscription_state
+            group_id (str): name of the consumer group to join for dynamic
+                partition assignment (if enabled), and to use for fetching and
+                committing offsets. Default: 'kafka-python-default-group'
```
Drizzt1991: default mismatch

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+            # select offsets that should be committed
+            offsets = {}
+            for partition in self._subscription.assigned_partitions():
+                tp = self._subscription.assignment[partition]
+                if tp.position != tp.committed and tp.has_valid_position:
+                    offsets[partition] = OffsetAndMetadata(tp.position, '')
+
+            try:
+                yield from self.commit_offsets(offsets)
+            except Errors.KafkaError as error:
```
Drizzt1991: Do we not need to backoff here?

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+                if error.retriable:
```
Drizzt1991: How about we change this to something like:
```Python
if error.retriable and not self._closing.done():
```
So if we close and failed to commit - we will see it in logs

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+
+    @asyncio.coroutine
+    def close(self):
+        """Close the coordinator, leave the current group
+        and reset local generation/memberId."""
+        self._closing.set_result(None)
+        if self._auto_commit_task:
+            try:
+                yield from self._auto_commit_task
+            except asyncio.CancelledError:
```
Drizzt1991: We can't have a CancelledError here. We changed it to another paradigm

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+        self._enable_auto_commit = enable_auto_commit
+        self._auto_commit_interval_ms = auto_commit_interval_ms
+        self._assignors = assignors
+        self._subscription = subscription
+        self._partitions_per_topic = {}
+        self._cluster = client.cluster
+        self._cluster.request_update()
+        self._cluster.add_listener(self._handle_metadata_update)
+        self._auto_commit_task = None
+        self._closing = asyncio.Future(loop=loop)
```
Drizzt1991: Add a comment to signal, that we use it for fast closure, as waiters will be released on it

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+            # attempt any resending if the request fails or times out.
+            request = LeaveGroupRequest(self.group_id, self.member_id)
+            try:
+                yield from self._send_req(self.coordinator_id, request)
+            except Errors.KafkaError as err:
+                log.error("LeaveGroup request failed: %s", err)
+            else:
+                log.info("LeaveGroup request succeeded")
+
+        self.generation = OffsetCommitRequest.DEFAULT_GENERATION_ID
```
Drizzt1991: Can we reuse the coordinator after closure? I think it's better to create new one and just forget about those params.

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+    @asyncio.coroutine
+    def commit_offsets(self, offsets):
+        """Commit specific offsets asynchronously.
+
+        Arguments:
+            offsets (dict {TopicPartition: OffsetAndMetadata}): what to commit
+
+        Raises error on failure
+        """
+        assert all(map(lambda k: isinstance(k, TopicPartition), offsets))
```
Drizzt1991: I think it is enough to do this check on consumer. Coordinator is protected interface anyway

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+        if not offsets:
+            log.debug('No offsets to commit')
+            return True
+
+        if (yield from self.coordinator_unknown()):
+            raise Errors.GroupCoordinatorNotAvailableError()
+        node_id = self.coordinator_id
+
+        # create the offset commit request
+        offset_data = collections.defaultdict(dict)
```
Drizzt1991: Kinda strange to repack this. Lets just construct a read-to-use structure in this loop, rather than inline below

Drizzt1991:
All those are just small changes. As a whole LGTM. :ship:

aiokafka/group_coordinator.py
```diff
@@ -0,0 +1,847 @@
...
+from aiokafka import ensure_future
+
+log = logging.getLogger(__name__)
+
+
+class GroupCoordinator(object):
+    """
+    GroupCoordinator implements group management for single group member
+    by interacting with a designated Kafka broker (the coordinator). Group
+    semantics are provided by extending this class.  See ConsumerCoordinator
```
Drizzt1991: Remove the `ConsumerCoordinator` comment. We have 1 implementation

PR #22, opened 2016-03-14
refactored consumer 

fabregas:


README.rst
```diff
@@ -36,6 +37,39 @@ Example of AIOKafkaProducer usage:
         loop.close()
 
 
+AIOKafkaConsumer
+****************
+
+AIOKafkaConsumer is a high-level, asynchronous message consumer.
+It also interacts with the assigned kafka Group Coordinator nodeto allow multiple consumers to load balance consumption of topics (requires kafka >= 0.9.0.0).
```
Drizzt1991: `It interacts with the assigned kafka Group Coordinator node to allow multiple consumers to load balance consumption of topics (requires kafka >= 0.9.0.0).`

README.rst
```diff
@@ -36,6 +37,39 @@ Example of AIOKafkaProducer usage:
...
+                except KafkaError as err:
+                    print("error while consuming message: ", err)
+
+        loop = asyncio.get_event_loop()
+        consumer = AIOKafkaConsumer('topic1', 'topic2', loop=loop, bootstrap_servers='localhost:1234')
+        loop.run_until_complete(consumer.start())
+        asyncio.async(consume_task(consumer))
+        loop.run_until_complete(asyncio.sleep(5))
+        loop.run_until_complete(consumer.stop())
+        c_task.close()
```
Drizzt1991: what's c_task?

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+            cluster metadata. This does not have to be the full node list.
+            It just needs to have at least one broker that will respond to a
+            Metadata API Request. Default port is 9092. If no servers are
+            specified, will default to localhost:9092.
+        client_id (str): a name for this client. This string is passed in
+            each request to servers and can be used to identify specific
+            server-side log entries that correspond to this client. Also
+            submitted to GroupCoordinator for logging with respect to
+            consumer group administration. Default: 'aiokafka-{version}'
+        group_id (str or None): name of the consumer group to join for dynamic
```
Drizzt1991: I really wander why would we set some string as default. I think it is more rational to set `None` as default here...

tests/test_consumer.py
```diff
@@ -30,71 +40,103 @@ def assert_message_count(self, messages, num_messages):
 
     @asyncio.coroutine
     def consumer_factory(self, **kwargs):
-
-        kwargs.setdefault('auto_commit', True)
-        group = kwargs.pop('group', self.id().encode('utf-8'))
-        consumer = SimpleAIOConsumer(self.client, group, self.topic, **kwargs)
-        yield from consumer._connect()
+        kwargs.setdefault('enable_auto_commit', True)
+        kwargs.setdefault('auto_offset_reset', 'earliest')
```
Drizzt1991: Pass normal keywords in the function. This looks strange

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+            Default: 50.
+        auto_offset_reset (str): A policy for resetting offsets on
+            OffsetOutOfRange errors: 'earliest' will move to the oldest
+            available message, 'latest' will move to the most recent. Any
+            ofther value will raise the exception. Default: 'latest'.
+        enable_auto_commit (bool): If true the consumer's offset will be
+            periodically committed in the background. Default: True.
+        auto_commit_interval_ms (int): milliseconds between automatic
+            offset commits, if enable_auto_commit is True. Default: 5000.
+        default_offset_commit_callback (callable): called as
```
Drizzt1991: We don't have this one

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+            value must be set lower than session_timeout_ms, but typically
+            should be set no higher than 1/3 of that value. It can be
+            adjusted even lower to control the expected time for normal
+            rebalances. Default: 3000
+        session_timeout_ms (int): The timeout used to detect failures when
+            using Kafka's group managementment facilities. Default: 30000
+        consumer_timeout_ms (int): number of millisecond to throw a timeout
+            exception to the consumer if no message is available for
+            consumption. Default: -1 (dont throw exception)
+        api_version (str): specify which kafka API version to use.
```
Drizzt1991: Remove other versions mentions. Specify only 2 allowed versions

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+            trigger custom actions when a commit request completes.
+        check_crcs (bool): Automatically check the CRC32 of the records
+            consumed. This ensures no on-the-wire or on-disk corruption to
+            the messages occurred. This check adds some overhead, so it may
+            be disabled in cases seeking extreme performance. Default: True
+        metadata_max_age_ms (int): The period of time in milliseconds after
+            which we force a refresh of metadata even if we haven't seen any
+            partition leadership changes to proactively discover any new
+            brokers or partitions. Default: 300000
+        partition_assignment_strategy (list): List of objects to use to
```
Drizzt1991: I don't understand why this is a list in `python-kafka`. @fabregas can you explain?

Drizzt1991: Ok, so it's a feature of Kafka 0.9.0, that allows downtime-less upgrades. From docs:

> Rolling Upgrades: To support rolling upgrades without downtime, there are two cases to consider: 
Changes affecting subscription: the protocol directly supports differing subscriptions, so there is no need for special handling. Members will only be assigned partitions compatible with their subscription.
Assignment strategy changes: to support a change to the assignment strategy, new versions must enable support both for the old assignment strategy and the new one. The coordinator will choose the old assignment strategy until all members have been updated. Then it will choose the new strategy. This preference is implicit in the order of the strategies in the JoinGroup request.

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+        self._retry_backoff_ms = retry_backoff_ms
+        self._enable_auto_commit = enable_auto_commit
+        self._auto_commit_interval_ms = auto_commit_interval_ms
+        self._partition_assignment_strategy = partition_assignment_strategy
+        self._key_deserializer = key_deserializer
+        self._value_deserializer = value_deserializer
+        self._fetch_min_bytes = fetch_min_bytes
+        self._fetch_max_wait_ms = fetch_max_wait_ms
+        self._max_partition_fetch_bytes = max_partition_fetch_bytes
+        self._consumer_timeout = consumer_timeout_ms / 100
```
Drizzt1991: Are those miliseconds? So why do we devide by 100?

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+            Kafka's group management feature. Heartbeats are used to ensure
+            that the consumer's session stays active and to facilitate
+            rebalancing when new consumers join or leave the group. The
+            value must be set lower than session_timeout_ms, but typically
+            should be set no higher than 1/3 of that value. It can be
+            adjusted even lower to control the expected time for normal
+            rebalances. Default: 3000
+        session_timeout_ms (int): The timeout used to detect failures when
+            using Kafka's group managementment facilities. Default: 30000
+        consumer_timeout_ms (int): number of millisecond to throw a timeout
```
Drizzt1991: Where do we throw a Timeout error? As I see it's some polling interval.

aiokafka/consumer.py
```diff
@@ -1,375 +1,612 @@
...
+        """
+        Get one message from Kafka
+        If no new messages occured, this method will wait it
+
+        Returns:
+            instance of collections.namedtuple("ConsumerRecord",
+                ["topic", "partition", "offset", "key", "value"])
+        """
+        while True:
+            while not self._subscription.assigned_partitions():
```
Drizzt1991: This can be done by a normal future.

Drizzt1991:
The Fetcher looks strange. Why do we need to manually ask for next fetch in the iterator? It looks kinda strange.

Drizzt1991:
How about we change the interface to:
```
consumer.getone(*partitions)  # Get 1 message, optionally getting only from specified partitions
consumer.getmany(*partitions, timeout_ms=None)  # Get a bulk of messages (wait for timeout if no messages in partition), optionally limit to specified partitions.
```
This way we can remove the pause() and resume() calls from public API, as those aren't that good in asyncio style.

Drizzt1991:
About Python35 interface, I think we can do this later, when we understand how the interface should look better.

aiokafka/consumer.py
Drizzt1991: List of objects to use to distribute partition ownership amongst consumer instances when group management is used. This preference is implicit in the order of the strategies in the list.
When assignment strategy changes: to support a change to the assignment strategy, new versions must enable support both for the old assignment strategy and the new one. The coordinator will choose the old assignment strategy until all members have been updated. Then it will choose the new strategy.

aiokafka/consumer.py
fabregas: done

aiokafka/consumer.py
fabregas: done

aiokafka/consumer.py
```diff
@@ -1,375 +1,608 @@
...
+        self._retry_backoff_ms = retry_backoff_ms
+        self._enable_auto_commit = enable_auto_commit
+        self._auto_commit_interval_ms = auto_commit_interval_ms
+        self._partition_assignment_strategy = partition_assignment_strategy
+        self._key_deserializer = key_deserializer
+        self._value_deserializer = value_deserializer
+        self._fetch_min_bytes = fetch_min_bytes
+        self._fetch_max_wait_ms = fetch_max_wait_ms
+        self._max_partition_fetch_bytes = max_partition_fetch_bytes
+        self._consumer_timeout = consumer_timeout_ms / 1000.
```
TarasLevelUp: Why is there a dot at the end? =) Divide is always float in Py3

aiokafka/consumer.py
```diff
@@ -1,375 +1,608 @@
...
+        self._fetch_min_bytes = fetch_min_bytes
+        self._fetch_max_wait_ms = fetch_max_wait_ms
+        self._max_partition_fetch_bytes = max_partition_fetch_bytes
+        self._consumer_timeout = consumer_timeout_ms / 1000
+        self._check_crcs = check_crcs
+        self._subscription = SubscriptionState(auto_offset_reset)
+        self._fetcher = None
+        self._coordinator = None
+        self._closed = False
+        self._iterator = None
```
Drizzt1991: Can't find this anywhere

aiokafka/consumer.py
```diff
@@ -1,375 +1,608 @@
...
-    def _get_message(self, get_partition_info=None, update_offset=True):
-        """
-        If no messages can be fetched, returns None.
-        If get_partition_info is None, it defaults to self.partition_info
-        If get_partition_info is True, returns (partition, message)
-        If get_partition_info is False, returns message
+    def getmany(self, *partitions, timeout_ms=0):
+        """Fetch data from assigned topics / partitions.
+
+        Records are fetched and returned in batches by topic-partition.
```
Drizzt1991: This is outdated info.

README.rst
```diff
@@ -36,6 +37,39 @@ Example of AIOKafkaProducer usage:
...
+
+        import asyncio
+        from kafka.common import KafkaError
+        from aiokafka import AIOKafkaConsumer
+
+        @asyncio.coroutine
+        def consume_task(consumer):
+            while True:
+                try:
+                    msg = yield from consumer.get_message()
```
Drizzt1991: Fix the example to use new interface

README.rst
```diff
@@ -36,6 +37,39 @@ Example of AIOKafkaProducer usage:
...
+                    print("consumed: ", msg.topic, msg.partition, msg.offset, msg.value)
+                except KafkaError as err:
+                    print("error while consuming message: ", err)
+
+        loop = asyncio.get_event_loop()
+        consumer = AIOKafkaConsumer('topic1', 'topic2', loop=loop, bootstrap_servers='localhost:1234')
+        loop.run_until_complete(consumer.start())
+        c_task = asyncio.async(consume_task(consumer))
+        loop.run_until_complete(asyncio.sleep(5))
```
Drizzt1991: Lets make it run until ctrl+C. It's better for an example

Drizzt1991:
I would still enhance the Fetcher, but this is good for first runs. Code style is OK. LGTM :ship:

TarasLevelUp:
Ok, on second view. Lets rewrite the Fetcher. I don't like it.

aiokafka/fetcher.py
```diff
@@ -0,0 +1,579 @@
...
+    def close(self):
+        if self._fetch_task:
+            self._fetch_task.cancel()
+            try:
+                yield from self._fetch_task
+            except asyncio.CancelledError:
+                pass
+            self._fetch_task = None
+
+    def _init_fetches(self):
```
TarasLevelUp: Remove this functions. You can just have a long-running background task for this. It will know better when to get new data

aiokafka/fetcher.py
```diff
@@ -0,0 +1,579 @@
...
+                        elif msg.offset != position:
+                            continue
+
+                        self._subscriptions.assignment[tp].position += 1
+                        yield msg
+                else:
+                    # these records aren't next in line based on the last
+                    # consumed position, ignore them they must be from
+                    # an obsolete request
+                    log.debug("Ignoring fetched records for %s at offset %s",
```
Drizzt1991: This is a strange error. Just how can we get outdated records here?

aiokafka/fetcher.py
```diff
@@ -0,0 +1,579 @@
...
+                raise Errors.InvalidMessageError(msg)
+            elif msg.is_compressed():
+                for record in self._unpack_message_set(tp, msg.decompress()):
+                    yield record
+            else:
+                key, value = self._deserialize(msg)
+                yield ConsumerRecord(
+                    tp.topic, tp.partition, offset, key, value)
+
+    def _message_generator(self, partitions):
```
Drizzt1991: Pls remove this generator. It's adding complexity, which is most certainlty not needed here. Just add a function, that will post process a single message.

aiokafka/fetcher.py
```diff
@@ -0,0 +1,579 @@
...
+
+    def _init_fetches(self):
+        """Send FetchRequests asynchronously for all assigned partitions.
+        """
+        if self._fetch_task is None:
+            self._fetch_task = ensure_future(
+                self._fetch_requests_routine(), loop=self._loop)
+
+    @asyncio.coroutine
+    def _fetch_requests_routine(self):
```
TarasLevelUp: This must be an endless background task, that will poll data for each assigned partition in a long-polling manner. We will poll for data only if:
* There are no data already ready for this partition (we have already prefetched a page for that partition)
* There are no pending requests for this partition
Basically we need to change the `wait` call to resume on first completed. If we have data for all partitions prefetched - set a waiter future, that will be resolved after:
* some partition cleared it's buffer
* rebalanced assignements

aiokafka/fetcher.py
```diff
@@ -0,0 +1,579 @@
...
+                        self._subscriptions.assignment[tp].position += 1
+                        yield msg
+                else:
+                    # these records aren't next in line based on the last
+                    # consumed position, ignore them they must be from
+                    # an obsolete request
+                    log.debug("Ignoring fetched records for %s at offset %s",
+                              tp, fetch_offset)
+
+    def next_record(self, partitions):
```
Drizzt1991: This needs some kind of traversing algorithm. Currently it's implemented in the iterator, but I think it's behaviour is not obvious (for example what will happen if we try to add an element to `tp_records` while it's in process of iteration?). 

What is needed is that we'd always process partitions equaly. How about:
* We maintain an ordered dict instead of a simple dict?
* Set a waiting future if no records prefetched. It may be released at each successful fetch request (if we waited for other partitions - just set another waiter).

aiokafka/fetcher.py
```diff
@@ -0,0 +1,579 @@
...
+
+            if not self._iterator:
+                self._iterator = self._message_generator(partitions)
+            try:
+                return next(self._iterator)
+            except StopIteration:
+                self._iterator = None
+                continue
+
+    def fetched_records(self, partitions):
```
Drizzt1991: This will be simpler - just return anything, that is contained in `self._records`. If we have no data - set waiter like with 1 item.

aiokafka/fetcher.py
TarasLevelUp: We can use subscription listener for this. Lets remove it from client's API. If needed we can add a more flexible event scheme.

TarasLevelUp:
Ok, i'm satisfied for now. Gives latency around 30-40ms to produce and consume 1 msg, stable. Rebalancing on consumer failure still needs to be tested, though.

PR #25, opened 2016-04-04
Producer batches

Drizzt1991:


TarasLevelUp:
I already reviewed the implementation. Looks good.
There is 1 problem, thou, we have to do 2 `yield from` calls for ACK:
```Python
send_future = yield from producer.send('foobar', b'test data')
result = yield from send_future
```
Why did we left it like this?
* `send()` call may block if we have too many undelivered data (reached max_batch_size). If we cancel during `send()` call message will never be delivered.
* The returned future, on the contrary, is a shielded result, as we already message for sending. Delivery will not be failed by cancellation, but we can still fail to deliver the message if we reach timeout.

Those two cases should probably be handled separately for exceptions.

Drizzt1991:
Just added a simple `send_and_wait` helper in the end.

PR #29, opened 2016-04-11
fixed fetcher fetched_records/next_record concurrency issue

fabregas:


popravich:
LGTM

aiokafka/fetcher.py
```diff
@@ -601,7 +601,8 @@ def next_record(self, partitions):
                 self._notify(self._wait_consume_future)
                 res_or_error.check_raise()
         # No messages ready. Wait for some to arrive
-        self._wait_empty_future = asyncio.Future(loop=self._loop)
+        if self._wait_empty_future is None or self._wait_empty_future.done():
+            self._wait_empty_future = asyncio.Future(loop=self._loop)
         yield from self._wait_empty_future
```
TarasLevelUp: Hmm, I think you probably want to shield this call.

TarasLevelUp: If many calls block it would be bad to cancel them all if 1 is cancelled.

PR #33, opened 2016-04-27
added some consumer checks

fabregas:
- check that subscribed topic/partition known
- check that send future is not done (can be cancelled)

### DIFF
--- a/aiokafka/consumer.py
+++ b/aiokafka/consumer.py
@@ -1,7 +1,8 @@
 import asyncio
 import logging
 
-from kafka.common import OffsetAndMetadata, TopicPartition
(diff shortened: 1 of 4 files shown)


aiokafka/consumer.py
```diff
@@ -209,8 +210,10 @@ def start(self):
             # using manual partitions assignment by topic(s)
             yield from self._client.force_metadata_update()
             partitions = []
-            for topic in self._topics:
```
TarasLevelUp: We don't need this variable anymore I think

fabregas: removed _topics attr

PR #48, opened 2016-09-19
Fix/multiple interfaces

JacobJakub:
While it iterates over all network interfaces, only the first one is used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import sys
 from setuptools import setup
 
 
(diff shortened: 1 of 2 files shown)


JacobJakub:
Hello, Can this be merged?

asvetlov:
@Drizzt1991 ?

Drizzt1991:
Thanks, I don't know how we missed that.

PR #54, opened 2016-10-03
Upgrade kafka python

dwoz:
Upgrade to work with newer versions of kafka-python


dwoz:
The master branch has all of these changes.

PR #56, opened 2016-10-14
Prevent committing offset when group coordinator is unavailable

### LINKED ISSUE #44: Endless "Skipping heartbeat: no active group:" error messages
If I let aiokafka connect to Kafka 0.10 after it has just started, it keeps spamming out these error messages and won't stop:

> [2016-07-21 08:00:04,646] ERROR [aiokafka.group_coordinator]: Skipping heartbeat: no active group: 
> [2016-07-21 08:00:04,747] ERROR [aiokafka.group_coordinator]: Skipping heartbeat: no active group: 
> [2016-07-21 08:00:04,849] ERROR [aiokafka.group_coordinator]: Skipping heartbeat: no active group: 
> [2016-07-21 08:00:04,950] ERROR [aiokafka.group_coordinator]: Skipping heartbeat: no active group:

If I stop my app, keep Kafka running and start my app again it will connect just fine.

This forces me to introduce `sleep()` commands to my code and hope that after an arbitrary delay everything works.

Lawouach:
Prevent committing offset when coordinator is not available

Fix #44 by not attempting to commit offsets when
we are not joined to a group.

This avoids looping forever and permits the group to be
joined again normally once it is available.

### DIFF
--- a/aiokafka/group_coordinator.py
+++ b/aiokafka/group_coordinator.py
@@ -212,7 +212,8 @@ class GroupCoordinator(object):
         self._subscription.mark_for_reassignment()
 
         # commit offsets prior to rebalance if auto-commit enabled
-        yield from self._maybe_auto_commit_offsets_sync()
+        if not (yield from self.coordinator_unknown()):
+            yield from self._maybe_auto_commit_offsets_sync()
 
         # execute the user's callback before rebalance
         log.debug("Revoking previously assigned partitions %s",


Drizzt1991:
Ok seems good for the autocommit problem, but I think we have a bigger problem here if it actually loops. Will look into it today.

Drizzt1991:
I do not like, that we lose the `commit` upon reconnection I do think we can process that last autocommit correctly after reconnection, will wrack my brain for that.

Drizzt1991:
Closing in favor of #57

Issue #58, opened 2016-10-16
Bump python-kafka to 1.3.1


Drizzt1991:
There are a few broken tests on this version

```
    @classmethod
    def encode(cls, items):
        # RecordAccumulator encodes messagesets internally
        if isinstance(items, io.BytesIO):
            size = Int32.decode(items)
            # rewind and return all the bytes
            items.seek(-4, 1)
            return items.read(size + 4)

        encoded_values = []
>       for (offset, message) in items:
E       ValueError: too many values to unpack (expected 2)

```

```
>       self.assertEqual(sorted(brokers), sorted(list(c_brokers)))
E       AssertionError: Lists differ: [(0, 'broker_1', 4567), (1, 'broker_2', 5678)] != [BrokerMetadata(nodeId=0, host='broker_1',[83 chars]one)]
E       
E       First differing element 0:
E       (0, 'broker_1', 4567)
E       BrokerMetadata(nodeId=0, host='broker_1', port=4567, rack=None)
E       
E       - [(0, 'broker_1', 4567), (1, 'broker_2', 5678)]
E       + [BrokerMetadata(nodeId=0, host='broker_1', port=4567, rack=None),
E       +  BrokerMetadata(nodeId=1, host='broker_2', port=5678, rack=None)]
```

rmuslimov:
+1 here. kafka-python 1.3.1 has important bugfixes. However, aiokafka is not working with latest kafka-python.

Drizzt1991:
Fixed in #60

PR #60, opened 2016-10-25
Kafka0.10 support

fabregas:
Please, review my implementation of "kafka 0.10 support" issue (#58)


Drizzt1991:
Wow, the desciption is a bit lacking, but GJ. Let me describe the change set here, pls confirm if I am correct:
* Bumped python-kafka version to 1.3.1
* Fixed `auto` version detection, to correctly handle 0.10.0.0 version rather than 0.9
* Changed lz4 compression, as in Kafka 0.10.0.0 it was changed due to [KIP-57](https://cwiki.apache.org/confluence/display/KAFKA/KIP-57+-+Interoperable+LZ4+Framing)
* Minor refactorings

aiokafka/fetcher.py
```diff
@@ -170,6 +172,12 @@ def __init__(self, client, subscriptions, *, loop,
         self._wait_consume_future = None
         self._wait_empty_future = None
 
+        if type(client.api_version) is tuple and client.api_version >= (0, 10):
```
Drizzt1991: Ok, I checked how ``python-kafka`` handles it, and bam - the same way as here, a shame rly. Well for now let's go with this, will try to poke ppl at python-kafka to do it properly later...

aiokafka/client.py
```diff
@@ -121,6 +138,14 @@ def bootstrap(self):
             raise ConnectionError(
                 'Unable to bootstrap from {}'.format(self.hosts))
 
+        # detect api version if need
+        if self._api_version == 'auto':
+            self._api_version = yield from self.check_version()
+        if type(self._api_version) is not tuple:
+            self._api_version = tuple(
+                map(int, self._api_version.split('.')))
+        self._version_id = 0 if self._api_version < (0, 10) else 1
```
Drizzt1991: Lets rename this to `self._api_v10` or something...

Drizzt1991:
Ok, I really wanted to use the Version API properly, but as ``python-kafka`` does the same for now it will do. Lets just change that name and good to go

Drizzt1991:
Cool

Issue #32, opened 2016-04-27
Add SSL support


TarasLevelUp:

Drizzt1991:
The kerberos will require a lot more work. If needed should be another issue.
As for SSL, the implementation is quite simple, but how should we test SSL support? We have docker too, so probably it should be upgraded to support ssl.

popravich:
If you only need a ssl proxy, you can try something like this: https://github.com/aio-libs/aioredis/commit/cd3c0eb2b50c99ecf66dd94150d339cf161f9854

Drizzt1991:
fixed by #81 SASL will be in a different issue

PR #81, opened 2016-11-17
Add ssl support

Drizzt1991:


tests/conftest.py
```diff
@@ -21,6 +24,36 @@ def docker():
     return libdocker.Client(version='auto')
 
 
+@pytest.yield_fixture(scope='session')
+def ssl_folder(docker_ip_address):
```
fabregas: maybe we can generate and store (commit) ssl cert/key in tests/data for example... and use it in tests

tests/conftest.py
```diff
@@ -57,27 +90,42 @@ def session_id():
...
 @pytest.yield_fixture(scope='session')
-def kafka_server(request, docker, docker_ip_address, unused_port, session_id):
+def kafka_server(request, docker, docker_ip_address,
+                 unused_port, session_id, ssl_folder):
     image = request.config.getoption('--docker-image')
-    docker.pull(image)
+    # Don't pull image if it available localy. Helps when developing new docker
+    # builds
+    if not docker.images(image):
+        docker.pull(image)
```
fabregas: I have some pygo/kafka:2.11_0.9.0.1 image locally (pulled on previous `make test` in master).
If you (or somebody else) changes this image on docker hub, I'll never pull it (with this code).

For testing new docker builds you can use --docker-image option in pytest.

tests/conftest.py
```diff
@@ -98,14 +146,29 @@ def loop(request):
...
+def setup_test_class(request, loop, kafka_server, ssl_folder):
     request.cls.loop = loop
-    request.cls.kafka_host, request.cls.kafka_port = kafka_server
+    khost, kport, ksslport = kafka_server
+    request.cls.kafka_host = khost
+    request.cls.kafka_port = kport
+    request.cls.kafka_ssl_port = ksslport
+    request.cls.ssl_folder = ssl_folder
+
+    def _create_context(self):
```
fabregas: why so implicit?) you pass `ssl_folder` to test class. So can implement this helper method in `KafkaIntegrationTestCase` class.

Drizzt1991:
Fixes SSL part of #32

PR #94, opened 2016-12-14
Minor bug fixes

fabregas:
fixed:
* fetcher.next_record() recursion issue
* client.check_version() issue for 0.10 version detection
* conn._closed_fut reinit issue (in reconnect case)


Drizzt1991:
Great work!

PR #96, opened 2016-12-14
wait topic creation on producer.send

fabregas:


Drizzt1991:
Wow, so MetadataRequest creates the topic... I thought it would be created on first send =(

PR #98, opened 2016-12-30
wait subscribed topic(s) creation in consumer on start

fabregas:
If topic does not exists in Kafka, consumer is crashes on start.

> $ python examples/simple_consumer.py 
  File "examples/simple_consumer.py", line 17, in <module>
    loop.run_until_complete(consume())
  File "/usr/lib64/python3.5/asyncio/base_events.py", line 387, in run_until_complete
    return future.result()
  File "/usr/lib64/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/lib64/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "examples/simple_consumer.py", line 10, in consume
    await consumer.start()
  File "/home/kst/.virtualenvs/aiokafka/lib/python3.5/site-packages/aiokafka/consumer.py", line 210, in start
    raise UnknownTopicOrPartitionError()

_wait_on_metadata() coroutine is moved to client and used both producer and consumer.

### DIFF
--- a/aiokafka/consumer.py
+++ b/aiokafka/consumer.py
@@ -157,6 +157,7 @@ class AIOKafkaConsumer(object):
             loop=loop, bootstrap_servers=bootstrap_servers,
             client_id=client_id, metadata_max_age_ms=metadata_max_age_ms,
             request_timeout_ms=request_timeout_ms,
+            retry_backoff_ms=retry_backoff_ms,
             api_version=api_version,
             ssl_context=ssl_context,
             security_protocol=security_protocol)
@@ -191,6 +192,7 @@ class AIOKafkaConsumer(object):
     @asyncio.coroutine
     def start(self):
         yield from self._client.bootstrap()
+        yield from self._wait_topics()
 
         if self._client.api_version < (0, 9):
             raise ValueError("Unsupported Kafka version: {}".format(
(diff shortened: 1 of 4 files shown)


tests/test_consumer.py
```diff
@@ -542,3 +540,20 @@ def test_consumer_group_without_subscription(self):
         yield from consumer.start()
         yield from asyncio.sleep(0.2, loop=self.loop)
         yield from consumer.stop()
+
+    @run_until_complete
+    def test_consumer_wait_topic(self):
+        topic = "some-test-topic-for-autocreate"
+        consumer = AIOKafkaConsumer(
+            topic, loop=self.loop, bootstrap_servers=self.hosts)
+        yield from consumer.start()
```
Drizzt1991: I think this test should:
* Start consumer
* start a background `getone` task.
* wait ~0.5 seconds. Just to be sure `getone` does not fail
* produce 1 message
* check message

aiokafka/consumer.py
```diff
@@ -191,6 +192,7 @@ def __init__(self, *topics, loop,
     @asyncio.coroutine
     def start(self):
         yield from self._client.bootstrap()
+        yield from self._wait_topics()
```
Drizzt1991: Why do we do it before we start coordinator and fetcher? This will sadly not fix:
```python
consumer = await AIOKafkaConsumer(bootstrap_servers="localhost:9092")
consumer.start()
consumer.subscribe("some_autocreate_topic")
```

Drizzt1991:
How about we change the subscribe to a coroutine, that will wait for metadata update?

tests/test_consumer.py
fabregas: good point! done

Drizzt1991:
well this is still better than what we have now. Will just merge, later we still need to do something with `subscribe` being a function.

Issue #88, opened 2016-12-05
Race condition in heartbeat


Artimi:
During testing our code we've tweaked some parameters of `AIOKafkaConsumer`. One of them is `heartbeat_interval_ms` that we set to 100 ms. After this we run a test that was waiting 100 ms between subscribing to more topics. Strange thing occurred sometimes (at first 1 of 20 cases, then 1 of 200 cases) that when I subscribed and unsubscribed a topic too fast `kafka/aiokafka` raised this exception:
```
    raise self._exception
/usr/lib/python3.5/asyncio/tasks.py:239: in _step
    result = coro.send(None)
messaging_client/mqclient/kafka_consumer.py:227: in stop
    await self._consumer.stop()
../../.virtualenvs/messaging/lib/python3.5/site-packages/aiokafka/consumer.py:279: in stop
    yield from self._coordinator.close()
../../.virtualenvs/messaging/lib/python3.5/site-packages/aiokafka/group_coordinator.py:132: in close
    yield from self.heartbeat_task
/usr/lib/python3.5/asyncio/futures.py:363: in __iter__
    return self.result()  # May raise too.
/usr/lib/python3.5/asyncio/futures.py:274: in result
    raise self._exception
/usr/lib/python3.5/asyncio/tasks.py:239: in _step
    result = coro.send(None)
../../.virtualenvs/messaging/lib/python3.5/site-packages/aiokafka/group_coordinator.py:789: in _heartbeat_task_routine
    yield from self.ensure_active_group()
../../.virtualenvs/messaging/lib/python3.5/site-packages/aiokafka/group_coordinator.py:593: in ensure_active_group
    yield from self._perform_group_join()
../../.virtualenvs/messaging/lib/python3.5/site-packages/aiokafka/group_coordinator.py:677: in _perform_group_join
    self.protocol, member_assignment_bytes)
../../.virtualenvs/messaging/lib/python3.5/site-packages/aiokafka/group_coordinator.py:266: in _on_join_complete
    self._subscription.assign_from_subscribed(assignment.partitions())
[...]

PR #106, opened 2017-02-10
Fast fix for issue #88. Coordinator race condition

Drizzt1991:
Relates to #88.


Drizzt1991:
The fix is not good enough. It will not fix issue with adding topics to subscribtion, as no conflict will occur.

Issue #46, opened 2016-08-05
New topics doesn't appear when consumer use pettern.


amelius0712:
AIOKafkaClient store list of topics and pass it into MetadataRequest. So if a new topic appear in kafka, 
then consumer did not know anything about it. Does it really need - to store topic list inside client? Or we just can always do MetadataRequest with empty topic list to get all inforamtion about current state of kafka?

Drizzt1991:
We used the `python-kafka` driver for reference there. At that time it did the same behaviour ( basically we can not subscribe to topics created after the client bootstrap without full client reconnect), so we decided to stick with it, rather than trying to do a better job there.

amelius0712:
Ок, but how can i get to know about new topics to make full reconnect?

Drizzt1991:
Ok, seems we do have a difference with ``kafka-python`` here. If metadata is updated it should catch the freshly created topic, but it does not. Ref: http://stackoverflow.com/questions/36153783/kafka-consumer-to-dynamically-detect-topics-added
@amelius0712 Kafka has no API for monitoring topic creation, but lowering metadata polling (controlled by ``metadata_max_age_ms``) should allow faster topic discovery. Thou currently it does not work for some reason.
Will patch in the next release.

Drizzt1991:
With #111 there should be improvement with this, at least subscribed topics will update if you use `group_id`

Drizzt1991:
Should be fully fixed in #143. Now consumer will see new topics after loading new cluster metadata even if `group_id is None`.

Drizzt1991:
Should be fixed as of 0.2.2. Sorry for long wait =)

PR #111, opened 2017-02-14
Add `exclude_internal_topics` option and fixed pattern subscription

### LINKED ISSUE #74: Add exclude.internal.topics option

Drizzt1991:
Fixes #74


Drizzt1991:
Should also resolve #46 issue in case of `group_id is not None`.

PR #115, opened 2017-02-17
Added fix to support compacted topics data, where some offsets can be…

Drizzt1991:
… skipped.


Drizzt1991:
This needs a bit better testing, but I would still like to have it in 0.2.1 release.

Drizzt1991:
Should probably split in into 2 separate PR's

PR #119, opened 2017-02-20
Make force_metadata update always update correctly. 

Drizzt1991:
It could wait fo outdated future before


Drizzt1991:
Part of #120

Issue #144, opened 2017-03-20
Reconnect to broker


Artimi:
Hi,
we are having a few network interruption lately and so I'm digging in reconnecting. I was not able to find reconnect feature in `aiokafka` in `conn.py` nor in `client.py`. Is it implented in `aiokafka` elsewhere or is it missing? 
I was looking in the kafka documentation and there is an option `reconnect.backoff.ms` for consumer and producer (https://kafka.apache.org/0102/documentation/#newconsumerconfigs) that prevents too fast reconnecting in case of failure. In `kafka-python` it is implemented in `KafkaClient._maybe_refresh_metadata` and another part in `KafkaClient._boostrap`.

I was trying to drop connection using [tcpkill](https://en.wikipedia.org/wiki/Tcpkill) and I discovered that it blocks connection from rising sequence of ports 
```
10.107.20.205:39614 > 172.17.0.2:9092: R 2713530037:2713530037(0) win 0
...
10.107.20.205:39616 > 172.17.0.2:9092: R 3723147408:3723147408(0) win 0
...
10.107.20.205:39618 > 172.17.0.2:9092: R 830983828:830983828(0) win 0
```
Do you have an idea how can this happen? Is there some reconnect after all? However when I stopped `tcpkill` no new messages went through although in `kafka-python` it reconnected and started to accepting messages again.

I'm willing to implement this feature but I would welcome any advice where it would be best to put (I was thinking about putting it in `conn.py` in case of some error in sending message).

Drizzt1991:
Hey there, that's new, it should just reconnect. How you actually reproduce this?

To answer your question of where it reconnects:
* https://github.com/aio-libs/aiokafka/blob/master/aiokafka/client.py#L309 - here we check if our connection is still `.connected()`, if not we reconnect.
* https://github.com/aio-libs/aiokafka/blob/master/aiokafka/conn.py#L180 - Here's where the connection actually notices that it's disconnected.

Artimi:
Ah, I'm sorry for wrong conclusion. It wasn't working for me because I didn't catch exceptions from `get()` and `commit()`. It seems that `aiokafka` is working flawlessly. Only thing that confused me is missing `reconnect.backoff.ms`. I was searching for that in the codebase and didn't find it so I concluded that `aiokafka` is not reconnecting. Sorry for my mistake and thanks for your response.

Drizzt1991:
@Artimi NP). Glad it was resolved.
[...]

PR #149, opened 2017-04-05
Reconnect on TimeoutError

Artimi:
This should solve https://github.com/aio-libs/aiokafka/issues/144. When there is raised exception `asyncio.TimeoutError` from `AIOKafkaConnection` in `Client.send` it will delete connection from `Client._conns`. This means that next time we do some request to this node new connection is created and therefore we get rid of stale connection. There is still raised `KafkaTimeoutError` so you can retry your request if needed. I added a simple test for this. Please advise if there is something wrong with my code.

### DIFF
--- a/tests/test_conn.py
+++ b/tests/test_conn.py
@@ -204,3 +204,12 @@ class ConnIntegrationTest(KafkaIntegrationTestCase):
         with self.assertRaises(ConnectionError):
             yield from conn.send(request)
         self.assertEqual(conn.connected(), False)
+
+    @run_until_complete
+    def test_close_disconnects_connection(self):
+        asyncio.set_event_loop(self.loop)
+        host, port = self.kafka_host, self.kafka_port
+        conn = yield from create_conn(host, port)
+        self.assertTrue(conn.connected())
+        conn.close()
+        self.assertFalse(conn.connected())
--- a/aiokafka/client.py
+++ b/aiokafka/client.py
@@ -376,6 +376,8 @@ class AIOKafkaClient:
         try:
             result = yield from future
         except asyncio.TimeoutError:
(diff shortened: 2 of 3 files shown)


Drizzt1991:
Thanks, few thoughts:

 * You need to close the socket before discarding connections, or you'll end up with stale sockets. I would recommend you just close the connection, rather than discarding it.
 * There could be multiple requests pipelined on this connection. Closing will probably raise ConnectionError in those, rather than KafkaTimeoutError, need to think if that's ok.

Drizzt1991:
To check yourself for coverage, please run:
```
make cov
make diff-cov
```
It will show you how much of your diff is covered by tests.

Artimi:
Thanks for feedback. I changed it to close and added test that `close()` really make connection `not conn.connected()`. Problem with tests was that I omitted `@run_until_complete` decorator. Yes if I close connection with multiple requests in line it will raise `ConnectionError` but I guess this is right because I will have new connection then and requests might be retried.

tests/test_conn.py
```diff
@@ -204,3 +204,12 @@ def invoke_osserror(*a, **kw):
         with self.assertRaises(ConnectionError):
             yield from conn.send(request)
         self.assertEqual(conn.connected(), False)
+
+    @run_until_complete
+    def test_close_disconnects_connection(self):
+        asyncio.set_event_loop(self.loop)
```
Drizzt1991: Don't do this in tests. It's not without a reason it's there, you probably didn't pass loop explicitly somewhere.

Artimi: Ok sorry about that I was just expanding test `test_global_loop_for_create_conn` and this remained there.

Issue #145, opened 2017-03-22
Missing requirement for tests in README.rst


Artimi:
In `README.rst` there is written that all you need to do is:
1. Install `docker`
2. Install `libsnappy-dev`
3. run `make setup`
4. run `make test`

However after few hours I've discovered that you need to have installed also utility `keytool` https://docs.oracle.com/javase/6/docs/technotes/tools/solaris/keytool.html that is part of Java (not installed on my computer before). It is used in script `gen-ssl-certs.sh` to generate few files and one of them was `br_server.keystore.jks`. `gen-ssl-certs.sh` on tests start silently failed and didn't make those files. Kafka in docker then failed everytime it tried to start on exception:
```
org.apache.kafka.common.KafkaException: org.apache.kafka.common.KafkaException: java.io.FileNotFoundException: /ssl_cert/br_server.keystore.jks (No such file or directory)
```
I recommend to update README with `keytool` requirements and also raise exception in fixture `ssl_folder.py` if `gen-ssl-certs.sh` does not return 0.

Drizzt1991:
Yea, took the `gen-ssl-certs.sh` from libdrkafka, kinda assumed it only uses openssl. Will think of how to better fix this. You can't actually start current docker image without valid certificates.

PR #172, opened 2017-05-24
Fix positions

arpanshah29:
self._update_fetch_positions expects a list. Function does not pass it in the correct format.

### DIFF
--- a/aiokafka/consumer.py
+++ b/aiokafka/consumer.py
@@ -469,7 +469,7 @@ class AIOKafkaConsumer(object):
             'Partition is not assigned'
         offset = self._subscription.assignment[partition].position
         if offset is None:
-            yield from self._update_fetch_positions(partition)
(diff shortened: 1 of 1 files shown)


Drizzt1991:
Ok, now this is a strange miss. Can you add a test for this change? Something simple like:
* Produce 1 msg to `self.topic` in ConsumerIntegration
* Create consumer without any subscribed topics
* Try to fetch offsets for `the self.topic`'s partition

arpanshah29:
Sounds good, how do I get the test environment up? Does it need a running docker container to be able to run tests?

Drizzt1991:
Great, you can start by reading https://github.com/aio-libs/aiokafka/tree/master#running-tests.
Also you will need `keytool` from java, I'm pretty sure it's installed with any jdk, but I may be wrong. For details see https://github.com/aio-libs/aiokafka/issues/145.

PR #179, opened 2017-07-07
Change subscription called only if necessary

Artimi:
This is subsequent PR to https://github.com/dpkp/kafka-python/pull/1132 . When we are subscribing by pattern mostly the set of topics is not changed but `change_subscription` is called anyway and then report warning that nothing happened. I don't thing this is message worthy of warning level. In `kafka-python` we agreed on that `change_subscription` should not be called if we can easily check that. 
Also I fixed a few typos I found in docstrings.


tvoinarovskyi:
Looks good 👍

PR #180, opened 2017-07-10
Reset buffer to 0 offset before retrying batch

tvoinarovskyi:


tvoinarovskyi:
@ngavrysh Hey, basically this is you're fix in https://github.com/Tubular/aiokafka, just put the `seek(0)` in `get_data_buffer()` function.

Issue #159, opened 2017-04-19
Fix the subscribe `listener` API


tvoinarovskyi:
One of the idea's of `listener`'s API is to allow setUp and tearDown of topic or partition resources. Like invalidate or populate caches, commit last batch of messages, etc.
In the current implementation we only allow to call **functions** `on_partitions_revoked` and `on_partitions_assigned`. It will not stall group management (new messages will be fetched right away, Join Group will happen right away) defeating the purpose.

TODO:
* Instead of just calling callbacks, check if `on_partitions_assigned` and `on_partitions_revoked` are by any chance coroutine functions (``asyncio.iscoroutinefunction``) and await them if they are.
* Add an example of usage to ``consumer.rst`` docs. Fix ``subscribe`` docstring.

PR #183, opened 2017-07-12
Allow on_partition_assigned and on_partition_revoked to be async functions

ask:
This is a backward compatible change, since it makes async optional.

We require these to be async functions since we need to call async code from within
the on_partitions_assigned and on_partitions_revoked handlers.

### DIFF
--- a/aiokafka/consumer.py
+++ b/aiokafka/consumer.py
@@ -472,7 +472,7 @@ class AIOKafkaConsumer(object):
             'Partition is not assigned'
         offset = self._subscription.assignment[partition].position
         if offset is None:
-            yield from self._update_fetch_positions([partition])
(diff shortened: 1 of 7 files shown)


tvoinarovskyi:
This does not work on still supported old python3's. Could you change it to [asyncio.iscoroutine](https://docs.python.org/3/library/asyncio-task.html#asyncio.iscoroutine)

tvoinarovskyi:
And yea, thanks for the work. It's quite a needed feature.Fixes #159

ask:
I messed it up it seems, not sure what's going on, it's suddenly showing wild differences between upstream

ask:
#190 replaces this pull request

PR #190, opened 2017-07-24
Allow on_partition_assigned and on_partition_revoked to be async functions

### LINKED ISSUE #159: Fix the subscribe `listener` API

ask:
This is a backward compatible change, since it makes async optional.

We require these to be async functions since we need to call async code from within
the on_partitions_assigned and on_partitions_revoked handlers.

Note: Fixes #159

### DIFF
--- a/aiokafka/consumer.py
+++ b/aiokafka/consumer.py
@@ -472,7 +472,7 @@ class AIOKafkaConsumer(object):
             'Partition is not assigned'
         offset = self._subscription.assignment[partition].position
         if offset is None:
-            yield from self._update_fetch_positions([partition])
+            yield from self._update_fetch_positions(partition)
             offset = self._subscription.assignment[partition].position
         return offset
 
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
Cool, looks good!

Issue #137, opened 2017-03-11
Consumer and Coordinator should use a separate socket for connections


tvoinarovskyi:
Followup dev ticket for #128. See also https://issues.apache.org/jira/browse/KAFKA-1925

tvoinarovskyi:
Fixed in #196

PR #196, opened 2017-07-28
Added separate socket groups to client

tvoinarovskyi:
Before we only created a single socket per node, but as protocol is synchronous socket can be blocked by requests like long poll fetch request (say for 500ms). It's bad as commits (sent to Coordinator which may be the same node) may be blocked by previous long-poll request. This PR adds noun of socket groups, so we can only have 1 socket to node in a group, but COORDINATION will be a separate group.

### DIFF
--- a/aiokafka/client.py
+++ b/aiokafka/client.py
@@ -24,6 +24,12 @@ __all__ = ['AIOKafkaClient']
 log = logging.getLogger('aiokafka')
 
 
+class ConnectionGroup:
+
+    DEFAULT = 0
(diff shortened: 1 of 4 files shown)


tvoinarovskyi:
@jettify @asvetlov Say did you need something similar? When we need 2 sockets to node. Any ideas how the interface should look. I'm a bit troubled by the noun `socket group`, maybe a better suggestion?

tvoinarovskyi:
Fixes #137

PR #209, opened 2017-08-22
Add flush method to AIOKafkaProducer

vineet-rh:
This separates out the logic to await flushing the message accumulator into a `flush()` method that can be used to flush remaining buffered messages without having to `close()` the producer.

### DIFF
--- a/aiokafka/producer.py
+++ b/aiokafka/producer.py
@@ -201,13 +201,17 @@ class AIOKafkaProducer(object):
         log.debug("Kafka producer started")
 
     @asyncio.coroutine
+    def flush(self):
+        """Wait untill all batches are Delivered and futures resolved"""
+        yield from self._message_accumulator.close()
+
+    @asyncio.coroutine
     def stop(self):
         """Flush all pending data and close all connections to kafka cluster"""
         if self._closed:
             return
 
-        # Wait untill all batches are Delivered and futures resolved
-        yield from self._message_accumulator.close()
+        yield from self.flush()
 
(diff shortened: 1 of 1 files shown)


aiokafka/message_accumulator.py
```diff
@@ -181,12 +181,18 @@ def set_api_version(self, api_version):
         self._api_version = api_version
 
     @asyncio.coroutine
-    def close(self):
-        self._closed = True
+    def flush(self):
         # NOTE: we copy to avoid mutation during `yield from` below
+        self._closed = True
```
tvoinarovskyi: Why do you set `self._closed` flag here? It's not needed I think.

aiokafka/producer.py
```diff
@@ -201,12 +201,16 @@ def start(self):
         log.debug("Kafka producer started")
 
     @asyncio.coroutine
+    def flush(self):
+        """Wait untill all batches are Delivered and futures resolved"""
```
tvoinarovskyi: *untill -> until please.

tvoinarovskyi:
Hey there, I would love to merge it, but the build it breakes the behaviour of ProducerClosed error if `send` called after `close`.

aiokafka/message_accumulator.py
vineet-rh: The idea was to avoid buffering more messages while flushing but now that I think of it that might not be needed.

tvoinarovskyi:
Thanks

PR #215, opened 2017-09-09
Add batch-oriented base Producer class

shargan:
The message-oriented model of AIOKafkaProducer is user-friendly and requires little knowledge of the library's innards to use effectively. It also limits performance by delegating all batch controls to the MessageAccumulator, which is hidden entirely from user-interaction, and creating futures for every single message sent.

This diff makes the following major changes:
- Leaves the AIOKafkaProducer interface intact. All existing use-cases should continue to work without modification.
- Adds AIOKafkaBaseProducer as a batch-oriented parent class with the primary interfaces of `create_batch()`, `get_batch()`, and `send_batch()`. Users of these interfaces operate on MessageBatch objects.
- Eliminates MessageAccumulator entirely. Its logic has been pulled into AIOKafkaBaseProducer with closer proximity to `sender_routine()`.

It also makes the following minor changes:
- Moves recognizes API versions into a class variable.
- Sets `self._closed = True` earlier in `producer.stop()` to ensure messages can't be added while awaiting other shutdown coroutines.
- Tweaks `MessageBatch.wait_drain()` to return an `asyncio.wait()` future like `MessageBatch.wait_deliver()`.
- Adds an `api_version` property mapping to `self.client.api_version`.
- Rewrites several `assert` statements as `if/raise` value checks.
- Pulls all existing tests from `tests/test_message_accumulator.py` into `tests/test_producer.py` with only minor updates.
- Eliminates the `_PRODUCER_CLIENT_ID_SEQUENCE` logic. Because the operation took place at the instance level, the class variable was never overwritten and every client simply ended up a value of `1`.

All tests currently succeed. A future PR will include additional tests specific to the new methods.

```
(.venv) aiokafka (shargan/flex-producer)$ make test FLAGS=--no-pull
extra=$(python -c "import sys;sys.stdout.write('--exclude tests/test_pep492.py') if sys.version_info[:3] < (3, 5, 0) else sys.stdout.write('')"); \
	flake8 aiokafka tests $extra
py.test -s --no-print-logs --docker-image aiolibs/kafka:2.11_0.10.1.1 --no-pull tests
======================================== test session starts ========================================
platform darwin -- Python 3.5.3, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
benchmark: 3.1.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/shargan/repos/aiokafka, inifile:
plugins: cov-2.5.1, catchlog-1.2.2, benchmark-3.1.1
collected 114 items

tests/test_client.py ................
tests/test_conn.py ..........
tests/test_consumer.py ..............................s......	...........x
tests/test_coordinator.py .............
tests/test_fetcher.py ......
tests/test_helpers.py s.
tests/test_pep492.py ....
tests/test_producer.py ..............

======================== 111 passed, 2 skipped, 1 xfailed in 113.30 seconds =========================
```


shargan:
One of the main motivations behind this change was performance. Using a very simple benchmarking tool, I compared production of 1,000,000 messages at 100 bytes each. All other settings were left at the default.

- `AIOKafkaProducer.send()`: 24782.04 msg/s
- `AIOKafkaProducer.send_and_wait()`: 1129.22 msg/s
- `AIOKafkaBaseProducer.send_batch()`: 51581.08 msg/s
- `AIOKafkaBaseProducer.send_batch_and_wait()`: 40135.60 msg/s

This sort of benchmark is meaningless in absolute terms, of course, but the relative difference is dramatic. My particular use case requires callbacks on message delivery, so moving from `send_and_wait()` to `send_batch_and_wait()` offers tremendous upside.

tvoinarovskyi:
For starters, if you have speed problems, open an issue and let's figure that together )

From what I see in your PR, you could have just added 2 methods `send_batch` and `send_batch_wait` to `Producer`, why did you go all the way to organize it as a base class? I can't just merge such a big change which basically moves code around. If there's something I missed, point it out.

As for your benchmark, I suspect you made a mistake there. `send` and `send_and_wait` are the same methods, there should not be any difference between those 2. You have to write the benchmark to send from several Tasks (like you would with several Threads in sync clients), that way you will have a CPU cap. What you did is probably just sending 1 msg at a time in a loop, which is wrong. You can post the benchmark and I'll point out where the difference in numbers come from and how to avoid it.

tvoinarovskyi:
So the question, apart from speed, did you need this change for anything else? Like do you have some use case when you need to operate on batches, rather than on single messages?

shargan:
> For starters, if you have speed problems, open an issue and let's figure that together )

Sure. Sorry for not doing that; my experience with other OSS projects has been "Send us code, not an issue."

> As for your benchmark, I suspect you made a mistake there. send and send_and_wait are the same methods, there should not be any difference between those 2. You have to write the benchmark to send from several Tasks (like you would with several Threads in sync clients), that way you will have a CPU cap. What you did is probably just sending 1 msg at a time in a loop, which is wrong. You can post the benchmark and I'll point out where the difference in numbers come from and how to avoid it.
> ...
> So the question, apart from speed, did you need this change for anything else? Like do you have some use case when you need to operate on batches, rather than on single messages?

My use case is routing messages between Kafka clusters with fair throughput (150-200k msg/s). Since consumer offsets are simply watermarks and the throughput is too high to maintain a map of attempted/delivered messages, ensuring message delivery before attempting the next is the only path I've found toward reliability (especially in events like unexpected router shutdowns).

So your suspicion of what I was benchmarking is accurate, but that was a deliberate approach rather than a mistake. Providing too little context is what happens when I submit a PR at the end of a long day.

This also touches on my need to operate on batches: obviously, waiting for an ack from the cluster before sending another message completely hamstrings my throughput. That's not a flaw in AIOKafka; it's just the logic of a networked system. Since Kafka writes messages in batches, I can get identical reliability by filling up a single batch and sending it. Doing so requires direct access to the batch itself.

> From what I see in your PR, you could have just added 2 methods send_batch and send_batch_wait to Producer, why did you go all the way to organize it as a base class?

I couldn't find a way to make send and send_batch play nicely, so I wanted to present them as two alternative APIs without duplicate logic. Producer's other user-friendly features (serialization and partitioning) are unnecessary in send_batch. I concluded that splitting the classes was the best way to avoid confusion. If you feel that docstring messages are enough, I'm happy to combine them back into one.

The other major change, eliminating MessageAccumulator, came about as I diagnosed race conditions in the send_batch logic. I had already separated the Producer classes, so adding a bit more backend logic to the base producer seemed natural.

shargan:
Created https://github.com/aio-libs/aiokafka/issues/216 to track the request for batch access.

shargan:
As discussed in Issue #216, I've submitted a lighter-weight version of this in https://github.com/aio-libs/aiokafka/pull/217.

Issue #216, opened 2017-09-10
Allow direct batch access in the Producer


shargan:
Spawned from https://github.com/aio-libs/aiokafka/pull/215

I have a use-case where my incoming stream cannot be advanced until messages are delivered. Doing so right now requires calling `await producer.send_and_wait(...)` in a loop which is terribly inefficient. Since Kafka writes entire MessageSets at once, having the ability to create, fill, and send batches directly would give me the fine control I need with far greater throughput.

The PR linked above is a functional first stab at this.

tvoinarovskyi:
Do I understand correctly, that you want to achieve:
```python
consumer = AIOKafkaConsumer(bootstrap_servers=consuming_cluster_servers)
producer = AIOKafkaProducer(bootstrap_servers=producing_cluster_servers)
consumer.subscribe(["topic1", "topic2"])

async for msg in consumer:
    await producer.send(value=msg.value, key=msg.key, timestamp=msg.timestamp)

```

tvoinarovskyi:
The code above will fetch messages as batches and send batch messages with high throughput. It will consume CPU to put messages from and in batch, but manually you will need to do the same basically...

tvoinarovskyi:
You can use a bit more efficient way to do that if you use `getmany()`:
```python
consumer = AIOKafkaConsumer(bootstrap_servers=consuming_cluster_servers)
producer = AIOKafkaProducer(bootstrap_servers=producing_cluster_servers)
consumer.subscribe(["topic1", "topic2"])

while True:
    mgsset = consumer.getmany(timeout_ms=1000)
    futs = []
    for tp, msgs in msgset.items():
        for msg in msgs:
            fut = await producer.send(value=msg.value, key=msg.key, timestamp=msg.timestamp)
            futs.append(fut)
    if futs:
        # This is basically where you will wait for the batch to be sent.
        await asyncio.gather(futs)
```

shargan:
That's an accurate description of my essential use case. Like most use-cases, it's very simplified (for example, I'm fetching from many topics and many consumers, routing varies by topic, acceptance policies vary per destination, and so on).

What you've described above is an improvement on awaiting each send_and_wait and similar to how I'm currently doing it (the send_and_wait example I gave above was a simplified explanation of my current logic). There are a few inefficiencies at work that make this less than ideal.

[...]

PR #217, opened 2017-09-13
Add lightweight batching interface to AIOKafkaProducer

shargan:
Lightweight batching interface (as opposed to heavier change in #215). Resolves issue https://github.com/aio-libs/aiokafka/issues/216.

This implementation offers callers a simple interface by which to create batches and submit them once they're populated:

```python
# examples/batch_produce.py
import asyncio
import random

from aiokafka.producer import AIOKafkaProducer

async def send_many(num):
    topic  = "my_topic"
    producer = AIOKafkaProducer(loop=loop)
    await producer.start()

    partitions = await producer.partitions_for(topic)
    part = random.choice(tuple(partitions))
    batch = producer.create_batch(topic, part)

    i = 0
    while i < num:
        payload = "Test message %d" % i
        fut = batch.append(None, payload.encode(), None)
        if fut is None:
            await producer.send_batch(batch)
            partitions = await producer.partitions_for(topic)
            part = random.choice(tuple(partitions))
            batch = producer.create_batch(topic, part)
            continue
        i += 1
    await producer.send_batch(batch)
    await producer.stop()


loop.run_until_complete(send_many(100))
loop.close()
```

Tests succeed (additional `make` arguments are only there to start docker properly on Mac OS):
```
(.venv) aiokafka (shargan/batch-producer)$ make test KAFKA_VERSION=0.10.1.1 FLAGS=--no-pull
extra=$(python -c "import sys;sys.stdout.write('--exclude tests/test_pep492.py') if sys.version_info[:3] < (3, 5, 0) else sys.stdout.write('')"); \
	flake8 aiokafka tests $extra
py.test -s --no-print-logs --docker-image aiolibs/kafka:2.11_0.10.1.1 --no-pull tests
========================================================= test session starts =========================================================
platform darwin -- Python 3.5.3, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
benchmark: 3.1.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/shargan/repos/aiokafka, inifile:
plugins: cov-2.5.1, catchlog-1.2.2, benchmark-3.1.1
collected 114 items

tests/test_client.py ................
tests/test_conn.py ..........
tests/test_consumer.py ..............................s.................x
tests/test_coordinator.py .............
tests/test_fetcher.py ......
tests/test_helpers.py s.
tests/test_message_accumulator.py ..
tests/test_pep492.py ....
tests/test_producer.py ............

========================================= 111 passed, 2 skipped, 1 xfailed in 112.68 seconds ==========================================
```

### DIFF
--- a/aiokafka/producer/message_accumulator.py
+++ b/aiokafka/producer/message_accumulator.py
@@ -18,7 +18,9 @@ class BatchBuilder:
     def __init__(self, magic, batch_size, compression_type):
         self._builder = LegacyRecordBatchBuilder(
             magic, compression_type, batch_size)
+        self._actual_size = 0
         self._relative_offset = 0
+        self._buffer = None
         self._closed = False
 
     def append(self, *, timestamp, key, value):
@@ -32,16 +34,28 @@ class BatchBuilder:
         if actual_size == 0:
             return 0
 
+        self._actual_size += actual_size
         self._relative_offset += 1
         return actual_size
 
-    def _build(self):
-        assert not self._closed
+    def close(self):
+        if self._closed:
+            return
         self._closed = True
-        buffer = self._builder.build()
+        data = self._builder.build()
+        self._buffer = io.BytesIO(Int32.encode(len(data)) + data)
         del self._builder
 
-        return io.BytesIO(Int32.encode(len(buffer)) + buffer)
+    def _build(self):
+        if not self._closed:
+            self.close()
+        return self._buffer
+
+    def size(self):
+        return self._actual_size
+
+    def __len__(self):
+        return self._relative_offset
 
 
 class MessageBatch:
@@ -56,6 +70,7 @@ class MessageBatch:
 
         # Waiters
         # Set when messages are delivered to Kafka based on ACK setting
+        self.future = create_future(loop)
         self._msg_futures = []
         # Set when sender takes this batch
         self._drain_waiter = create_future(loop=loop)
@@ -80,24 +95,29 @@ class MessageBatch:
     def done(self, base_offset=None, exception=None):
         """Resolve all pending futures"""
         for relative_offset, future in enumerate(self._msg_futures):
-            if future.done():
-                continue
-            if exception is not None:
-                future.set_exception(exception)
-            elif base_offset is None:
-                future.set_result(None)
-            else:
-                res = RecordMetadata(self._tp.topic, self._tp.partition,
-                                     self._tp, base_offset + relative_offset)
-                future.set_result(res)
+            self._set_future(future, base_offset, relative_offset, exception)
+        self._set_future(self.future, base_offset, 0, exception)
 
-    def wait_deliver(self):
-        """Wait until all message from this batch is processed"""
-        return asyncio.wait(self._msg_futures, loop=self._loop)
+    def _set_future(self, future, base_offset, relative_offset, exception):
+        if future.done():
+            return
+        if exception is not None:
+            future.set_exception(exception)
+        elif base_offset is None:
+            future.set_result(None)
+        else:
+            res = RecordMetadata(self._tp.topic, self._tp.partition,
+                                 self._tp, base_offset + relative_offset)
+            future.set_result(res)
 
-    def wait_drain(self):
+    def wait_deliver(self, timeout=None):
         """Wait until all message from this batch is processed"""
-        return self._drain_waiter
+        return asyncio.wait([self.future], timeout=timeout, loop=self._loop)
+
+    def wait_drain(self, timeout=None):
+        """Wait until all message from this batch is processed"""
+        return asyncio.wait(
+            [self._drain_waiter], timeout=timeout, loop=self._loop)
 
     def expired(self):
         """Check that batch is expired or not"""
@@ -159,17 +179,8 @@ class MessageAccumulator:
 
         pending_batches = self._batches.get(tp)
         if not pending_batches:
-            magic = 0 if self._api_version < (0, 10) else 1
-            builder = BatchBuilder(
-                magic, self._batch_size, self._compression_type
-            )
-            batch = MessageBatch(
-                tp, builder, self._batch_ttl, self._loop)
-            self._batches[tp].append(batch)
-
-            if not self._wait_data_future.done():
-                # Wakeup sender task if it waits for data
-                self._wait_data_future.set_result(None)
+            builder = self.create_builder()
+            batch = self.add_batch(builder, tp)
         else:
             batch = pending_batches[-1]
 
@@ -178,8 +189,7 @@ class MessageAccumulator:
             # Batch is full, can't append data atm,
             # waiting until batch per topic-partition is drained
             start = self._loop.time()
-            yield from asyncio.wait(
-                [batch.wait_drain()], timeout=timeout, loop=self._loop)
+            yield from batch.wait_drain(timeout)
             timeout -= self._loop.time() - start
             if timeout <= 0:
                 raise KafkaTimeoutError()
@@ -236,3 +246,34 @@ class MessageAccumulator:
         self._wait_data_future = create_future(loop=self._loop)
 
         return nodes, unknown_leaders_exist
+
+    def create_builder(self):
+        magic = 0 if self._api_version < (0, 10) else 1
+        return BatchBuilder(magic, self._batch_size, self._compression_type)
+
+    def add_batch(self, builder, tp):
+        """Add BatchBuilder to queue by topic-partition.
+
+        This method applies no queue limits, so batches may be enqueued up to
+        memory limits.
+
+        Arguments:
+            builder (BatchBuilder): batch object to enqueue.
+            tp (TopicPartition): topic and partition to enqueue this batch for.
+
+        Returns:
+            asyncio.Future: object that will be set when the batch is delivered
+                to the broker.
+
+        Raises:
+            aiokafka.errors.ProducerClosed: the accumulator has already been
+                closed and flushed.
+        """
+        if self._closed:
+            raise ProducerClosed()
+
+        batch = MessageBatch(tp, builder, self._batch_ttl, self._loop)
+        self._batches[tp].append(batch)
+        if not self._wait_data_future.done():
+            self._wait_data_future.set_result(None)
+        return batch
--- a/aiokafka/producer/message_accumulator.py
+++ b/aiokafka/producer/message_accumulator.py
@@ -55,7 +55,7 @@ class BatchBuilder:
         else:
             return self._builder.size()
 
-    def length(self):
+    def record_count(self):
         return self._relative_offset
 
 
@@ -271,7 +271,6 @@ class MessageAccumulator:
 
         Returns:
             MessageBatch: delivery wrapper around the BatchBuilder object.
-                to the broker.
 
         Raises:
             aiokafka.errors.ProducerClosed: the accumulator has already been
@@ -282,11 +281,11 @@ class MessageAccumulator:
         if self._closed:
             raise ProducerClosed()
 
+        start = self._loop.time()
         while timeout > 0:
-            start = self._loop.time()
             pending = self._batches.get(tp)
             if pending:
-                yield from pending[-1].wait_drain()
+                yield from pending[-1].wait_drain(timeout=timeout)
                 timeout -= self._loop.time() - start
             else:
                 return self._append_batch(builder, tp)
--- a/aiokafka/producer/message_accumulator.py
+++ b/aiokafka/producer/message_accumulator.py
@@ -18,7 +18,6 @@ class BatchBuilder:
     def __init__(self, magic, batch_size, compression_type):
         self._builder = LegacyRecordBatchBuilder(
             magic, compression_type, batch_size)
-        self._actual_size = 0
         self._relative_offset = 0
         self._buffer = None
         self._closed = False
@@ -34,7 +33,6 @@ class BatchBuilder:
         if actual_size == 0:
             return 0
 
-        self._actual_size += actual_size
         self._relative_offset += 1
         return actual_size
 
@@ -52,9 +50,12 @@ class BatchBuilder:
         return self._buffer
 
     def size(self):
-        return self._actual_size
+        if self._buffer:
+            return self._buffer.getbuffer().nbytes
+        else:
+            return self._builder.size()
 
-    def __len__(self):
+    def length(self):
         return self._relative_offset
 
 
@@ -180,7 +181,7 @@ class MessageAccumulator:
         pending_batches = self._batches.get(tp)
         if not pending_batches:
             builder = self.create_builder()
-            batch = self.add_batch(builder, tp)
+            batch = self._append_batch(builder, tp)
         else:
             batch = pending_batches[-1]
 
@@ -251,29 +252,42 @@ class MessageAccumulator:
         magic = 0 if self._api_version < (0, 10) else 1
         return BatchBuilder(magic, self._batch_size, self._compression_type)
 
-    def add_batch(self, builder, tp):
-        """Add BatchBuilder to queue by topic-partition.
-
-        This method applies no queue limits, so batches may be enqueued up to
-        memory limits.
-
-        Arguments:
-            builder (BatchBuilder): batch object to enqueue.
-            tp (TopicPartition): topic and partition to enqueue this batch for.
-
-        Returns:
-            asyncio.Future: object that will be set when the batch is delivered
-                to the broker.
-
-        Raises:
-            aiokafka.errors.ProducerClosed: the accumulator has already been
-                closed and flushed.
-        """
-        if self._closed:
-            raise ProducerClosed()
-
+    def _append_batch(self, builder, tp):
         batch = MessageBatch(tp, builder, self._batch_ttl, self._loop)
         self._batches[tp].append(batch)
         if not self._wait_data_future.done():
             self._wait_data_future.set_result(None)
         return batch
+
+    @asyncio.coroutine
+    def add_batch(self, builder, tp, timeout):
+        """Add BatchBuilder to queue by topic-partition.
+
+        Arguments:
+            builder (BatchBuilder): batch object to enqueue.
+            tp (TopicPartition): topic and partition to enqueue this batch for.
+            timeout (int): time in seconds to wait for a free slot in the batch
+                queue.
+
+        Returns:
+            MessageBatch: delivery wrapper around the BatchBuilder object.
+                to the broker.
+
+        Raises:
+            aiokafka.errors.ProducerClosed: the accumulator has already been
+                closed and flushed.
+            aiokafka.errors.KafkaTimeoutError: the batch could not be added
+                within the specified timeout.
+        """
+        if self._closed:
+            raise ProducerClosed()
+
+        while timeout > 0:
+            start = self._loop.time()
+            pending = self._batches.get(tp)
+            if pending:
+                yield from pending[-1].wait_drain()
+                timeout -= self._loop.time() - start
+            else:
+                return self._append_batch(builder, tp)
+        raise KafkaTimeoutError()
--- a/aiokafka/producer/producer.py
+++ b/aiokafka/producer/producer.py
@@ -528,16 +528,14 @@ class AIOKafkaProducer(object):
         """
         return self._message_accumulator.create_builder()
 
-    def send_batch(self, builder, topic, partition=None):
+    @asyncio.coroutine
+    def send_batch(self, batch, topic, *, partition):
         """Submit a BatchBuilder for publication.
 
         Arguments:
-            builder (BatchBuilder): batch object to be published.
+            batch (BatchBuilder): batch object to be published.
             topic (str): topic where the batch will be published.
-            partition (int, optional): optionally specify the partition to
-                where this batch will be published. If unset, the partition
-                will be selected using the configured `partitioner` with key
-                and value set to None.
+            partition (int): partition where this batch will be published.
 
         Returns:
             asyncio.Future: object that will be set when the batch is
@@ -546,5 +544,6 @@ class AIOKafkaProducer(object):
         partition = self._partition(topic, partition, None, None, None, None)
         tp = TopicPartition(topic, partition)
         log.debug("Sending batch to %s", tp)
-        batch = self._message_accumulator.add_batch(builder, tp)
-        return batch.future
+        msg_batch = yield from self._message_accumulator.add_batch(
+            batch, tp, self._request_timeout_ms / 1000)
+        return msg_batch.future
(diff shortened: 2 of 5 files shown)


aiokafka/producer.py
```diff
@@ -496,3 +496,39 @@ def _partition(self, topic, partition, key, value,
         available = list(self._metadata.available_partitions_for_topic(topic))
         return self._partitioner(
             serialized_key, all_partitions, available)
+
+    def create_batch(self, topic, partition=None):
+        """Create and return an empty MessageBatch.
+
+        The batch is not queued for send until submission to ``send_batch``.
+        Use of both ``create_batch`` and ``send`` together is discouraged as it
```
tvoinarovskyi: Just how are they unexpected? We should set a clear limit or allow both to be used.

aiokafka/producer.py
```diff
@@ -496,3 +496,39 @@ def _partition(self, topic, partition, key, value,
...
+        may lead to unexpected results.
+
+        Arguments:
+            topic (str): topic where the batch will be published
```
tvoinarovskyi: Why do we create the batch for some partition? I would like it to use the default partitioning strategy...

aiokafka/producer.py
```diff
@@ -496,3 +496,39 @@ def _partition(self, topic, partition, key, value,
...
+            partition (int, optional): optionally specify a partition. If not
+                set, the partition will be selected using the configured
+                'partitioner' with key and value set to None.
+
+        Returns:
+            MessageBatch: empty batch to be filled and submitted by the caller.
+        """
+        partition = self._partition(topic, partition, None, None, None, None)
+        tp = TopicPartition(topic, partition)
+        return self._message_accumulator.batch_factory(tp)
```
tvoinarovskyi: I don't think it's a good idea to use thus type for public Api. Lets just create a new, simple class, that will be used here. It can be an adapter around this one.

tvoinarovskyi: Also we need docs for Batch class methods.

aiokafka/producer.py
```diff
@@ -496,3 +496,39 @@ def _partition(self, topic, partition, key, value,
...
+
+        Arguments:
+            batch (MessageBatch): batch object to be published.
+
+        Returns:
+            asyncio.Future: object that will be set when the batch is
+                delivered.
+        """
+        log.debug("Sending batch to %s", batch.tp)
+        return self._message_accumulator.add_batch(batch)
```
tvoinarovskyi: Can this lead to message out of order problems? We have a guarantee on message order, i want to preserve it.

aiokafka/producer.py
```diff
@@ -496,3 +496,39 @@ def _partition(self, topic, partition, key, value,
...
+        may lead to unexpected results.
+
+        Arguments:
+            batch (MessageBatch): batch object to be published.
+
+        Returns:
+            asyncio.Future: object that will be set when the batch is
+                delivered.
+        """
+        log.debug("Sending batch to %s", batch.tp)
```
tvoinarovskyi: We need to add an explicit flag for debug messages. This is out of the scope of this PR, just a reminder.

tvoinarovskyi:
We also need doc changes to `producer.rst` and Api docs

aiokafka/producer.py
tvoinarovskyi: Ok, it's optional, saw that... This is a bit of a problem... How above moving those arguments to `send_batch`

aiokafka/message_accumulator.py
```diff
@@ -58,16 +61,20 @@ def _is_full(self, key, value):
...
         if value is not None:
             needed_bytes += len(value)
+        if needed_bytes > self._batch_size:
+            raise MessageSizeTooLargeError(
+                "Message is %d bytes which is larger than the maximum batch"
+                " size configured with max_batch_size" % needed_bytes)
         return self._buffer.tell() + needed_bytes > self._batch_size
 
-    def append(self, key, value, timestamp_ms):
+    def append(self, key, value, timestamp_ms, future=False):
```
tvoinarovskyi: Lets move the `future` to constructor, that way we will have a way to differentiate between batches.

aiokafka/message_accumulator.py
```diff
@@ -91,32 +98,40 @@ def append(self, key, value, timestamp_ms):
...
+
+    def _set_future(self, future, base_offset, relative_offset, exception):
+        if future.done():
+            return
+        if exception is not None:
+            future.set_exception(exception)
+        elif base_offset is None:
+            future.set_result(None)
+        else:
+            res = RecordMetadata(self.tp.topic, self.tp.partition,
```
tvoinarovskyi: I'm not sure how to treet RecordMetadata for batch resuts... Maybe a new class, like BatchMetadata...

aiokafka/producer.py
shargan: The batch object requires a TP for instantiation (the TP is later used to populate RecordMetadata objects). I can change that and set the TP during send/drain if you'd prefer.

I'm not sure what you mean by "use the default partitioning strategy" -- batches by their nature are sent to a single partition. The messages within a batch don't get repartitioned during send.

aiokafka/producer.py
shargan: As I said in my previous PR, this is why I originally separated the classes out: to make it clearer that these methods aren't very compatible.

They don't play nicely because `send` ends up appending a message to the first batch with space--creating one if it doesn't exist. When a batch is sent via this `send_batch`, it's simply added to the `MessageAccumulator` queue. If that batch isn't full, calls to `send` that end up at the same TP will be appended to that batch, meaning that offsets won't match what the original caller was doing.

Another potential interaction comes with calls in this order:
1. send -- a fresh batch is created and appended to the queue
2. send_batch -- a full batch is added to the queue
3. send -- now waits because the full batch is the "active" one, even though plenty of space remains in the batch created in step 1

There are a couple other edge conditions but these are the big ones.

aiokafka/producer.py
shargan: What worries you about using this class for a public API?

The goal of this change is to reduce indirection and offer an interface "closer to the metal" specifically to support more flexible use-cases and, eventually, swapping parts for increased performance (e.g., one of my long term goals is to be able to swap out the Message class in use for one optimized to my use case). I'd rather fix what needs to be fixed in MessageBatch for it to work as a public interface than add another layer of indirection.

aiokafka/producer.py
shargan: Agreed.

aiokafka/producer.py
shargan: Order internal to each method send and send_batch is correctly preserved. Order between those calls is not since send waits for the active batch to drain while send_batch happily adds a batch to the queue without regard for its current size.

aiokafka/message_accumulator.py
shargan: I'm not following this. Can you show me what you mean?

aiokafka/message_accumulator.py
shargan: Whatever you'd prefer is fine by me. The offset strategy I chose here (returning the offset of the first message written) matches the value returned in ProduceResponse, so I think the values are sensible even if the class name is not.

aiokafka/producer.py
tvoinarovskyi: > When a batch is sent via this send_batch, it's simply appended to the queue. If that batch isn't full, calls to send against the same TP will be appended to that batch, meaning that offsets won't match what the original batch caller was doing.

1) You need to close the batch for append it's submitted to `send_batch`. Ie. _is_full should return False.
2) Don't see why offsets would not match, as they are tracked in batch itself.

> Another potential interaction comes with calls in this order:

Maybe lets an explicit flag, when creating batches, that blocks `send` API until that batch is sent. Ie, every `create_batch` must match with `send_batch` calls. What do you think?

aiokafka/producer.py
tvoinarovskyi: Because that class is full of methods, that are not meant to be used, like `done`, `expired`, `wait_deliver`, `drain_ready` and so on. I want public API to be simple and not allow the user to shoot himself.

tvoinarovskyi: Second, the `ttl` will be counted incorrectly if you use this class.

aiokafka/producer.py
tvoinarovskyi: It requires TP because you used some internal class for it. If you create a separate one, it will be ok.

tvoinarovskyi: `use the default partitioning strategy` I wanted to use `partitioner`, but that requires the key of message... Most likely we will require explicit partition to be passed for batches

aiokafka/producer.py
tvoinarovskyi: I mean for example if 1 batch is sent and we add another batch, will those 2 be always sent in the proper order on retries

aiokafka/message_accumulator.py
tvoinarovskyi: Add a flag `future_per_message=True` to constructor and remove it from `append` kwargs.
Code will be:
```python
if self._future_per_message:
    future = loop.create_future()
    self._message_futures.append(future)
    return future
else:
    return self._future
```

aiokafka/message_accumulator.py
tvoinarovskyi: But `checksum`, `timestamp`, `serialized_key_size` and `serialized_value_size` don't really have any meaning thou.

tvoinarovskyi: Add a new class to `structs.py`, say BatchMetadata. It will contain `offset`, `topic`, `partition` and `timestamp` fields, where the timestamp is the one from the server.

aiokafka/producer.py
tvoinarovskyi: Split MessageBatch into 2 classes: `ProducerBatch` and `MessageBatch`. The second is not aware of topic and partition, will only contain the resulting formatted buffer. The first will be aware of `tp`, `ttl`, futures.

aiokafka/producer.py
shargan: Yes. The one-batch-inflight-at-a-time is regulated by sender_routine and _send_produce_req. Batches that need to be retried are reenqueued before the node is removed from inflight, so that same batch will be pulled off on the next sender_routine cycle.

aiokafka/message_accumulator.py
shargan: > But checksum, timestamp, serialized_key_size and serialized_value_size don't really have any meaning thou.

Are you sure you're not looking at ConsumerRecord? RecordMetadata has none of those fields.

> Add a new class to structs.py, say BatchMetadata. It will contain offset, topic, partition and timestamp fields, where the timestamp is the one from the server.

This is almost identical to RecordMetadata (except for the timestamp field):
```python
RecordMetadata = collections.namedtuple(
    'RecordMetadata', ['topic', 'partition', 'topic_partition', 'offset'])
```

Like I said, I'm happy to add a new class if that's your preference, but I think we're looking at two different pieces of code right now.

aiokafka/message_accumulator.py
shargan: Sorry, I should have been more explicit. Can you clarify what you meant by "that way we will have a way to differentiate between batches"?

aiokafka/producer.py
shargan: Okay.

aiokafka/producer.py
shargan: > 2. Don't see why offsets would not match, as they are tracked in batch itself.

In that case, a message would have been added to the batch without the caller's knowledge so their expectation for what was written is no longer accurate. Whether this is a problem depends on what the caller's doing, but it would violate the explicit control I'm looking for here.

> Maybe lets an explicit flag, when creating batches, that blocks send API until that batch is sent. Ie, every create_batch must match with send_batch calls. What do you think?

Sure.

aiokafka/message_accumulator.py
tvoinarovskyi: Currently, any batch is the same to sender routine, but if we add a flag like that we can see if it's an explicit user batch or one created from `send()` calls. Not that I can find any reason for that )

aiokafka/message_accumulator.py
tvoinarovskyi: Ooops =) Looked at `kafka-python` implementation. Hmm, it's a bug actually, RecordBatch should have a `timestamp`. Will create an issue.

shargan:
_For the sake of discussion, **InternalBatch** refers to the proposed internal-use batch with futures, ttls, drain methods, etc. and **ExternalBatch** refers to the API-use batch with the message buffer and append methods._

I spent the last two days trying to find a way to gracefully split MessageBatch into two separate classes as discussed and came up short. Each attempt resulted in duplicate logic and/or weird data sharing.

For example, `wait_drain` would presumably be an attribute on InternalBatch since it's needed internally but not externally. But it's only used in `add_message`, which naturally needs access to the message buffer. The drain process itself is an internally facing operation that needs access to the message buffer.
 
I worked for a while with the strategy of wrapping an ExternalBatch inside an InternalBatch, but ended up with two inextricable classes that were calling each other's methods and accessing each other's internal data. This dramatically reduced clarity--the exact opposite of what you're asking for.

So my most recent commit attempts to make MessageBatch a bit more suitable to both use cases. Specifically:
- All internal-use methods have been prefixed with an underscore to clearly indicate to users which are intended as API endpoints.
- It completely decouples TP from MessageBatch. This was easy since everywhere batch._tp was used, TP was already available from another source.
- It moves the encoding logic from `drain_ready()` into `close()`.
- The drain-specific logic has been moved from `drain_ready()` into `_drain()`.
- Batch ttl is now set via `set_tt()` rather than during initialization. `expired()` has been tweaked accordingly so it now returns the correct value for user-created batches.
- Methods `age()`, `size()`, and `__len__()` have been added to ensure callers have the information they need to interact with batches without poking at the internals.

The net result is, I hope, a scheme that's better suited for presenting as an API without constraining internal logic.

The addition of `close()` also means that the conflict between `send()` and `send_batch()` is now resolved. If the caller wishes complete control over the batch, they can call `batch.close()` before calling `producer.send_batch()`. Otherwise, the MessageAccumulator will freely add messages to that batch until it's drained.

If this looks acceptable to you, I'll flesh out the MessageBatch docstrings and producer.rst.

aiokafka/message_accumulator.py
tvoinarovskyi: Any update on moving this bool?

tvoinarovskyi:
Hey there, maybe I did not express myself enough when suggesting the split of MessageBatch into 2. When we need to compose the batch (after `create_batch`) we only need to properly create the buffer, no reference to Partition and no futures are required there. This will ease testing and when we'll do optimization. Futures are only needed after the batch is sent (ie. in `send_batch`).

To make it easier to understand I've split the class in a separate PR #220. Is it possible for you to rebase on top of it?

shargan:
I've rebased on master and submitted a new version of this diff using BatchBuilder as requested.

One notable change is that I've added the `BatchBuilder.close()` method so callers have the option of closing their builder before submitting (to ensure total control over the batch's contents) or not (to allow graceful concurrent use of both `producer.send()` and the batch interface).

examples/batch_produce.py
```diff
@@ -0,0 +1,33 @@
...
+    topic  = "my_topic"
+    producer = AIOKafkaProducer(loop=loop)
+    await producer.start()
+
+    batch = producer.create_batch()
+
+    i = 0
+    while i < num:
+        payload = "Test message %d" % i
+        if not batch.append(key=None, value=payload.encode(), timestamp=None):
```
tvoinarovskyi: Set explicit encoding for `encode()`. That way it will be self-documentary to what it does for people with less python knowledge.

examples/batch_produce.py
```diff
@@ -0,0 +1,33 @@
+import asyncio
+import random
+
+from aiokafka.producer import AIOKafkaProducer
+
+loop = asyncio.get_event_loop()
+
+
+async def send_many(num):
```
tvoinarovskyi: Apart from here, we need this example in docs.

aiokafka/producer/message_accumulator.py
```diff
@@ -32,16 +34,28 @@ def append(self, *, timestamp, key, value):
         if actual_size == 0:
             return 0
 
+        self._actual_size += actual_size
         self._relative_offset += 1
         return actual_size
 
-    def _build(self):
-        assert not self._closed
+    def close(self):
```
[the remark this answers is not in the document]
shargan: The benefit is graceful coexistence of both the batch interface and the direct `send()` interface without the need to artificially limit use cases. What's the downside?

aiokafka/producer/message_accumulator.py
```diff
@@ -236,3 +246,34 @@ def drain_by_nodes(self, ignore_nodes):
...
         return nodes, unknown_leaders_exist
+
+    def create_builder(self):
+        magic = 0 if self._api_version < (0, 10) else 1
+        return BatchBuilder(magic, self._batch_size, self._compression_type)
+
+    def add_batch(self, builder, tp):
+        """Add BatchBuilder to queue by topic-partition.
+
+        This method applies no queue limits, so batches may be enqueued up to
```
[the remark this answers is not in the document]
shargan: Since the batch interface is intended to grant the user greater control over the data flow and offers an optional flow control mechanism (wait for the batch to be delivered before enqueueing the next or don't), I reasoned that implicit flow control was an unnecessary complication. I'll add it back in.

aiokafka/producer/producer.py
```diff
@@ -517,3 +517,34 @@ def _partition(self, topic, partition, key, value,
...
+        """
+        return self._message_accumulator.create_builder()
+
+    def send_batch(self, builder, topic, partition=None):
+        """Submit a BatchBuilder for publication.
+
+        Arguments:
+            builder (BatchBuilder): batch object to be published.
+            topic (str): topic where the batch will be published.
+            partition (int, optional): optionally specify the partition to
```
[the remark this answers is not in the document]
shargan: Sure, but why make it keyword-only?

shargan:
I'll update documentation once the technical implementation is complete.

aiokafka/producer/message_accumulator.py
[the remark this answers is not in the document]
tvoinarovskyi: The downside is that you need to call `close()` to get the expected behavior.

aiokafka/producer/message_accumulator.py
```diff
@@ -35,13 +36,27 @@ def append(self, *, timestamp, key, value):
...
 
-        return io.BytesIO(Int32.encode(len(buffer)) + buffer)
+    def _build(self):
+        if not self._closed:
+            self.close()
+        return self._buffer
+
+    def size(self):
+        if self._buffer:
+            return self._buffer.getbuffer().nbytes
```
tvoinarovskyi: Why do you use such a strange way to get the size of the buffer? The creation of a memoryview is not so cheap. Why not call `self._buffer.tell()`? It should always be at the end anyway.

aiokafka/producer/message_accumulator.py
```diff
@@ -35,13 +36,27 @@ def append(self, *, timestamp, key, value):
...
+        else:
+            return self._builder.size()
+
+    def length(self):
```
tvoinarovskyi: No, `length` and `size` have a similar meaning and it's confusing. How about `record_count()`.

aiokafka/producer/message_accumulator.py
```diff
@@ -236,3 +247,47 @@ def drain_by_nodes(self, ignore_nodes):
...
+        Raises:
+            aiokafka.errors.ProducerClosed: the accumulator has already been
+                closed and flushed.
+            aiokafka.errors.KafkaTimeoutError: the batch could not be added
+                within the specified timeout.
+        """
+        if self._closed:
+            raise ProducerClosed()
+
+        while timeout > 0:
```
tvoinarovskyi: This is not a really good pattern for a timeout. It calls `loop.time()` 2ce and the time between cycles is not handled properly.

```python
start = self._loop.time()
remaining = timeout
while remaining > 0:
    pending = self._batches.get(tp)
    if pending:
        yield from pending[-1].wait_drain(timeout=remaining)
        remaining = timeout - (loop.time() - start)
    else:
        return self._append_batch(builder, tp)
```
This way we only call `time()` once.

aiokafka/producer/message_accumulator.py
```diff
@@ -236,3 +247,47 @@ def drain_by_nodes(self, ignore_nodes):
...
+            start = self._loop.time()
+            pending = self._batches.get(tp)
+            if pending:
+                yield from pending[-1].wait_drain()
```
tvoinarovskyi: Pass timeout to `wait_drain()`. See my snippet above, it already has the timeout passed.

aiokafka/producer/message_accumulator.py
```diff
@@ -236,3 +247,47 @@ def drain_by_nodes(self, ignore_nodes):
...
+
+        Arguments:
+            builder (BatchBuilder): batch object to enqueue.
+            tp (TopicPartition): topic and partition to enqueue this batch for.
+            timeout (int): time in seconds to wait for a free slot in the batch
+                queue.
+
+        Returns:
+            MessageBatch: delivery wrapper around the BatchBuilder object.
+                to the broker.
```
tvoinarovskyi: Probably a typo.

tvoinarovskyi:
No test to see if `add_batch` raises a KafkaTimeoutError.

aiokafka/producer/message_accumulator.py
shargan: No, it's not particularly cheap. That's why I originally had `BatchBuilder` tracking the size independently.

The buffer will only be at the end after the request is issued to the broker. During the time between `_build()` and the actual on-the-wire request creation in `producer._send_produce_req()` the buffer's position is 0. There are multiple yields to the event loop in that space--particularly if a batch is already in-flight.

The good news is that this method is user-facing and more likely to be called before submission of the batch than after, so most uses will use the cheaper `_builder.size()` call.

If you see this as an issue, we can save the size between calls after `_builder` is deleted so `getbuffer()` is only called once. Since this is a convenience method to the user whose use case is mostly pre-submission, I don't consider it a big deal.

aiokafka/producer/message_accumulator.py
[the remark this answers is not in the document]
shargan: Considering that batches are intended to give greater flexibility and finer control to the user, I don't see the downside as strongly as you do. How do you feel about adding a `close=True` kwarg to `producer.send_batch()` that governs this behavior?

tvoinarovskyi:
Cool, LGTM. Ship it %)

PR #225, opened 2017-09-25
Moved LegacyRecords from kafka-python's new protocol

tvoinarovskyi:


tvoinarovskyi:
it's much faster for PyPy and 30% faster for cPython

PR #226, opened 2017-09-25
Record builder c ext

tvoinarovskyi:


tvoinarovskyi:
Travis builds don't use Cext, something's not right there (

aiokafka/record/_legacy_records.pxd
```diff
@@ -0,0 +1,34 @@
...
+        char _magic
+        int _decompressed
+        LegacyRecord _main_record
+
+    @staticmethod
+    cdef inline _LegacyRecordBatchCython new(Py_buffer buf, char magic)
+
+    cdef int _decompress(self, char compression_type) except -1
+    cdef int64_t _read_last_offset(self) except -1
+    cdef inline _check_bounds(self, int pos, int size)
```
asvetlov: `void` return type?

PR #229, opened 2017-09-28
Send session_timeout_ms to GroupCoordinator constructor

shargan:
Consumer is currently swallowing the `session_timeout_ms` kwarg. This diff sends it on to the GroupCoordinator as presumably intended.


tvoinarovskyi:
Wow, now this is strange... Thanks

PR #231, opened 2017-09-28
Update TP position when a fetch request returns OffsetOutOfRangeError

shargan:
If `_fetch_requests_routine()` receives `OffsetOutOfRangeError` in response to a request the offending TP is marked as requiring a reset. However, `update_fetch_positions()`, the method responsible for performing the reset, is only called if the subscription changes or a user calls one of the consumer's seek methods. As a result, the fetcher retains ownership of the TP but no longer consumes from it.

This is especially problematic when processing data with `auto_offset_reset='earliest'` against a broker with specific retention times. Data may be pruned by the time autocommit kicks in and the TP never recovers.

The patch included here works but changes test behavior significantly. If this approach works for you, I'll update the tests accordingly.

### DIFF
--- a/aiokafka/consumer/fetcher.py
+++ b/aiokafka/consumer/fetcher.py
@@ -368,6 +368,7 @@ class Fetcher:
     @asyncio.coroutine
     def _proc_fetch_request(self, node_id, request):
         needs_wakeup = False
+        needs_position_update = []
         try:
             response = yield from self._client.send(node_id, request)
         except Errors.KafkaError as err:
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
I'll need to look closely into this one. We have some strange racing with update_fetch_position, not sure how to fix it too. Maybe a big change for this will be required.
Anyway I'm happy to merge this one if tests pass

tvoinarovskyi:
Ie. I'm ok with this fix, but probably the problem will require more attension

shargan:
I agree, that logic feels ready for a revision. I'll get the tests corrected for this diff to solve the immediate issue and may take a stab at a deeper change later.

shargan:
Tests added and functioning as expected.

Issue #218, opened 2017-09-13
Add `timestamp` field to RecordMetadata


tvoinarovskyi:
If broker uses LogAppendTime for `timestamp` we need to know what was set after the batch is added. So if we receive `log_append_time ` in Produce Response we will use that as timestamp, else we use the send time, or timestamp provided to `send()` call.

PR #237, opened 2017-10-15
Add timestamp to RecordMetadata

tvoinarovskyi:
* Added timestamp to RecordMetadata, returned from produce future


tvoinarovskyi:
Fixes #218

PR #239, opened 2017-10-16
Make the saved future in `send` weak reference for returned one

tvoinarovskyi:
Frankly not sure about this. It gives ~5% speedup when we don't read the future, but this actually breaks code like such:
```python
fut = await producer.send(topic, payload, partition=partition)
fut.add_done_callback(lambda fut: print(fut.result()))
fut = None
```
Because this will lose the reference we will never call `lambda fut: print(fut.result())`.

### DIFF
--- a/aiokafka/producer/message_accumulator.py
+++ b/aiokafka/producer/message_accumulator.py
@@ -1,6 +1,7 @@
 import asyncio
 import collections
 import io
+from weakref import ref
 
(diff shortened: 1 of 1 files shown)


vineetgoel:
Ideally we should make this optional. We do set a bunch of callbacks to messages sent async and the guarantees there would probably break with this.

tvoinarovskyi:
I dont see much benefit in this

PR #245, opened 2017-10-25
Performance improvements to _LegacyRecordBatchBuilderPy

shargan:
Highlights include:
- Perform `append()` type checking only when necessary (upon error)
- Wrap all `_encode_msg()` logic into a single `struct.pack_into()` call to reduce overhead
- Flatten multi-method invocations wherever possible without losing clarity
- Preallocate buffer space rather than expanding on each individual message
- Calculate key and value length only once

The results:
```
(.venv) aiokafka (master)$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 906 us +- 60 us
.....................
batch_append_v1: Mean +- std dev: 946 us +- 52 us
(.venv) aiokafka (master)$ python bench_records.py simple
1000000 messages in 7.83s (127758.34 msg/s)

(.venv) aiokafka (shargan/record-performance)$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 598 us +- 29 us
.....................
batch_append_v1: Mean +- std dev: 647 us +- 35 us
(.venv) aiokafka (shargan/record-performance)$ python bench_records.py simple
1000000 messages in 5.50s (181839.19 msg/s)
```

### DIFF
--- a/aiokafka/record/_legacy_records.pyx
+++ b/aiokafka/record/_legacy_records.pyx
@@ -21,7 +21,7 @@ cdef extern from "Python.h":
     object PyMemoryView_FromMemory(char *mem, ssize_t size, int flags)
 
 
-# This should be before _cutil to generate include for `winsock2.h` before 
+# This should be before _cutil to generate include for `winsock2.h` before
 # `windows.h`
 from aiokafka.record cimport _hton as hton
 from aiokafka.record cimport _cutil as cutil
@@ -360,6 +360,8 @@ cdef class _LegacyRecordBatchBuilderCython:
         cdef:
             Py_ssize_t pos
             Py_ssize_t size
+            Py_ssize_t key_size
+            Py_ssize_t value_size
             char *buf
             int64_t ts
             LegacyRecordMetadata metadata
@@ -372,9 +374,12 @@ cdef class _LegacyRecordBatchBuilderCython:
         else:
             ts = timestamp
 
+        key_size = _size_of_bytearray(key)
+        value_size = _size_of_bytearray(value)
+
         # Check if we have room for another message
         pos = PyByteArray_GET_SIZE(self._buffer)
-        size = _size_in_bytes(self._magic, key, value)
+        size = _size_in_bytes(self._magic, key_size, value_size)
         # We always allow at least one record to be appended
         if offset != 0 and pos + size >= self._batch_size:
             return None
@@ -398,10 +403,10 @@ cdef class _LegacyRecordBatchBuilderCython:
 
     # Size calculations. Just copied Java's implementation
 
-    def size_in_bytes(self, offset, timestamp, key, value):
+    def size_in_bytes(self, offset, timestamp, key_size, value_size):
         """ Actual size of message to add
         """
-        return _size_in_bytes(self._magic, key, value)
+        return _size_in_bytes(self._magic, key_size, value_size)
 
     @staticmethod
     def record_overhead(char magic):
@@ -415,6 +420,7 @@ cdef class _LegacyRecordBatchBuilderCython:
             object compressed
             char *buf
             Py_ssize_t size
+            Py_ssize_t value_size
             uint32_t crc
 
         if self._compression_type != 0:
@@ -429,7 +435,9 @@ cdef class _LegacyRecordBatchBuilderCython:
                     compressed = lz4_encode(bytes(self._buffer))
             else:
                 return 0
-            size = _size_in_bytes(self._magic, key=None, value=compressed)
+            value_size = _size_of_bytearray(compressed)
+            size = _size_in_bytes(
+                self._magic, key_size=0, value_size=value_size)
             # We will just write the result into the same memory space.
             PyByteArray_Resize(self._buffer, size)
 
@@ -447,32 +455,28 @@ cdef class _LegacyRecordBatchBuilderCython:
         return self._buffer
 
 
-cdef Py_ssize_t _size_in_bytes(char magic, object key, object value) except -1:
+cdef Py_ssize_t _size_in_bytes(char magic, Py_ssize_t key_size, Py_ssize_t value_size) except -1:
     """ Actual size of message to add
+    """
+    if magic == 0:
+        return LOG_OVERHEAD + RECORD_OVERHEAD_V0_DEF + key_size + value_size
+    else:
+        return LOG_OVERHEAD + RECORD_OVERHEAD_V1_DEF + key_size + value_size
+
+
+cdef Py_ssize_t _size_of_bytearray(object obj) except -1:
+    """ Properly return size of bytearray object
     """
     cdef:
         Py_buffer buf
-        Py_ssize_t key_len
-        Py_ssize_t value_len
 
-    if key is None:
-        key_len = 0
+    if obj is None:
+        return 0
     else:
-        PyObject_GetBuffer(key, &buf, PyBUF_SIMPLE)
-        key_len = buf.len
+        PyObject_GetBuffer(obj, &buf, PyBUF_SIMPLE)
+        obj_size = buf.len
         PyBuffer_Release(&buf)
-
-    if value is None:
-        value_len = 0
-    else:
-        PyObject_GetBuffer(value, &buf, PyBUF_SIMPLE)
-        value_len = buf.len
-        PyBuffer_Release(&buf)
-
-    if magic == 0:
-        return LOG_OVERHEAD + RECORD_OVERHEAD_V0_DEF + key_len + value_len
-    else:
-        return LOG_OVERHEAD + RECORD_OVERHEAD_V1_DEF + key_len + value_len
+        return obj_size
 
 
 cdef int _encode_msg(
--- a/aiokafka/record/_legacy_records.pyx
+++ b/aiokafka/record/_legacy_records.pyx
@@ -348,11 +348,12 @@ cdef class _LegacyRecordBatchBuilderCython:
     CODEC_LZ4 = ATTR_CODEC_LZ4
 
     def __cinit__(self, char magic, char compression_type,
-                  Py_ssize_t batch_size):
+                  Py_ssize_t batch_size, Py_ssize_t buffer_chunk_size=0):
         self._magic = magic
         self._compression_type = compression_type
         self._batch_size = batch_size
         self._buffer = bytearray()
+        # ignore buffer_chunk_size
 
     def append(self, int64_t offset, timestamp, key, value):
         """ Append message to batch.
@@ -403,17 +404,21 @@ cdef class _LegacyRecordBatchBuilderCython:
 
     # Size calculations. Just copied Java's implementation
 
-    def size_in_bytes(self, offset, timestamp, key_size, value_size):
+    def size_in_bytes(self, offset, timestamp, key, value):
         """ Actual size of message to add
         """
+        key_size = _size_of_bytearray(key)
+        value_size = _size_of_bytearray(value)
         return _size_in_bytes(self._magic, key_size, value_size)
 
     @staticmethod
     def record_overhead(char magic):
         if magic == 0:
             return RECORD_OVERHEAD_V0_DEF
-        else:
+        elif magic == 1:
             return RECORD_OVERHEAD_V1_DEF
+        else:
+            raise ValueError("Unsupported magic: %d" % magic)
 
     cdef int _maybe_compress(self) except -1:
         cdef:
--- a/aiokafka/record/legacy_records.py
+++ b/aiokafka/record/legacy_records.py
@@ -412,7 +412,8 @@ class _LegacyRecordBatchBuilderPy(LegacyRecordBase):
     def build(self):
         """Compress batch to be ready for send"""
         self._maybe_compress()
-        return memoryview(self._buffer)[:self._pos]
+        del self._buffer[self._pos:]
+        return self._buffer
 
     def size(self):
         """ Return current size of data written to buffer
--- a/aiokafka/record/legacy_records.py
+++ b/aiokafka/record/legacy_records.py
@@ -361,7 +361,8 @@ class _LegacyRecordBatchBuilderPy(LegacyRecordBase):
                 "%ds"  # value => bytes
                 % (key_size, value_size),
                 buf, 0, offset, length, 0, magic, attributes,
-                key_size or -1, key or b"", value_size or -1, value or b"")
+                key_size if key is not None else -1, key or b"",
+                value_size if value is not None else -1, value or b"")
         else:
             length += self.KEY_OFFSET_V1
             struct.pack_into(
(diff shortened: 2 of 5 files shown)


shargan:
bench_records.py is a stupid-simple micro-benchmark I was using mostly for profiling:
```python
import random
import sys
import time

from aiokafka.producer.message_accumulator import BatchBuilder

BATCH_SIZE = 1024 * 128
MAGIC = 1

def bench_simple(num, payload):
    magic = MAGIC
    size = BATCH_SIZE

    batch = BatchBuilder(MAGIC, BATCH_SIZE, None)
    i = 0
    while i < num:
        if batch.append(timestamp=None, key=None, value=payload) is None:
            batch._build()
            batch = BatchBuilder(MAGIC, BATCH_SIZE, None)
        else:
            i += 1

def main(arg):
    funcs = {"simple": bench_simple}
    num = 1000000
    size = 100
    payload = ("m" * size).encode()

    bench = funcs[arg]
    start = time.time()
    bench(num, payload)
    delta = time.time() - start
    print("%d messages in %.2fs (%.2f msg/s)" % (num, delta, num / delta))

if __name__ == "__main__":
    main(sys.argv[1])
```

Profiling master vs. these changes shows the substance of these gains. Here's the most interesting values from each:
```
VERSION    FUNC              CALLS      TOTTIME     CUMTIME
master     append()          1001022    3.145       10.812
PR         append()          1001022    2.204       6.164

master     _encode_msg()     1000000    3.696       5.400
PR         _encode_msg()     1000000    2.192       3.127

master     size_in_bytes()   1001022    0.552       1.449
PR         size_in_bytes()   1001022    0.342       0.342

master     calc_crc32()      1000000    0.349       0.606
PR         binascii.crc32()  1000000    0.241       0.241
```

tvoinarovskyi:
Hmmm, not sure about all of the changes. For now could you make the build pass, it broke.

aiokafka/record/legacy_records.py
```diff
@@ -278,155 +285,154 @@ def __repr__(self):
...
+
+    def _encode_msg(self, start_pos, size, offset, timestamp, key_size, key,
+                    value_size, value, attributes=0):
         """ Encode msg data into the `msg_buffer`, which should be allocated
             to at least the size of this message.
         """
         magic = self._magic
-        buf = self._buffer
-        pos = start_pos
+        buf = memoryview(self._buffer)[start_pos:]
```
tvoinarovskyi: Why do we create a memoryview here? We can set the position to write from in pack_into call.

aiokafka/record/legacy_records.py
```diff
@@ -11,6 +11,9 @@
 )
 
 
+NoneType = type(None)
```
tvoinarovskyi: Where is it used?

aiokafka/record/legacy_records.py
```diff
@@ -278,155 +285,154 @@ def __repr__(self):
 class _LegacyRecordBatchBuilderPy(LegacyRecordBase):
 
     def __init__(self, magic, compression_type, batch_size):
+        assert magic in [0, 1]
         self._magic = magic
         self._compression_type = compression_type
         self._batch_size = batch_size
-        self._buffer = bytearray()
+        self._buffer = bytearray(batch_size)
```
tvoinarovskyi: Not sure about this overalocation... We still do bounding checks anyway, does it give that much speed boost? The problem here is that it's a user specified field... It may be pretty big. This can chunk alot of memory away(

tvoinarovskyi: We could try to maintain a list of bytearrays instead and join on build. May be faster

aiokafka/record/legacy_records.py
```diff
@@ -278,155 +285,154 @@ def __repr__(self):
...
-        if offset != 0 and pos + size >= self._batch_size:
-            return None
-
-        # Allocate proper buffer length
-        self._buffer.extend(bytearray(size))
 
-        # Encode message
-        crc = self._encode_msg(pos, offset, timestamp, key, value)
+        # calculating length is not cheap; only do it once
+        key_size = len(key) if key is not None else 0
```
tvoinarovskyi: The size calculations and struct in a single call is good. GJ

aiokafka/record/legacy_records.py
```diff
@@ -278,155 +285,154 @@ def __repr__(self):
...
+
+        except struct.error:
+            # perform expensive type checking only to translate struct errors
+            # to human-readable messages
+            if type(offset) != int:
+                raise TypeError(offset)
+            if type(timestamp) != int:
+                raise TypeError(timestamp)
+            if not isinstance(key, (bytes, bytearray, memoryview, NoneType)):
+                raise TypeError("Unsupported type for key: %s" % type(key))
```
tvoinarovskyi: Not so sure about if this is a good idea. We perform len() call on those. It will raise a different error.

shargan:
Can you clarify what you find harder to read in this code? Personally, I find it quite a bit easier to read but that's a style-preference thing. The diff makes it look a lot uglier than it is. Is there something in particular you find hard to parse?

aiokafka/record/legacy_records.py
shargan: Lines 332 and 334. Those lines were previously broken into two:
```
if not (key is None or
    isinstance(key, (bytes, bytearray, memoryview))):
```

The use of NoneType (which for some reason isn't automatically in the global namespace) allows the expression to be condensed onto a single line:
```
if not isinstance(key, (bytes, bytearray, memoryview, NoneType)):
```

tvoinarovskyi: `is None` check is the fastest opcode in Python... The isinstance(NoneType) on the other hand is slow.

shargan: Right, but this diff only performs type checking in exceptional circumstances rather than on every `builder.append()`. Since the outcome of this block of code will always be a raised exception, I judged that the minor decrease in performance was worth the increased readability. I'm happy to switch it back if you'd prefer.

aiokafka/record/legacy_records.py
shargan: The performance difference is significant but not huge. There are two costs in extending the bytearray: creating the second bytearray and performing the actual extension. These costs grow with the size of each bytearray. Here's an illustration of the difference:

```python
>>> from timeit import timeit
>>> def oneshot():
...   return bytearray(16384)
...
>>> def multiple():
...  b = bytearray()
...  for _ in range (100):
...   b.extend(bytearray(150))
...
>>> timeit("oneshot()", setup="from __main__ import oneshot, multiple")
0.5214737450005487
>>> timeit("multiple()", setup="from __main__ import oneshot, multiple")
33.52806271798909
```

I don't think memory boundaries are a huge concern. First, each buffer can already take up the specified amount of memory. Second, unless the user is going through the batch interface (in which case they're explicitly taking control anyway) only two batches are allowed per TP. It would take fairly exceptional circumstances to cause noticeable memory pressure for the vast majority of running environments. Even at Netflix's scale (up to 10k partitions per cluster), the default batch settings would max out at ~320MB of memory.

I suspect most use-cases max out at less than 500 partitions, which by default would put this around 16MB of memory. If someone is tweaking their batch size, hopefully they realize that changes their memory use profile. Even if not, though, it's unlikely this will create problems.

tests/record/test_legacy.py
```diff
@@ -193,6 +193,7 @@ def test_read_log_append_time_v1():
 @pytest.mark.parametrize("magic", [0, 1])
 def test_reader_corrupt_record_v0_v1(magic):
     buffer = _make_compressed_batch(magic)
+    buffer = bytearray(buffer)
```
tvoinarovskyi: Why does it have to be a bytearray here?

aiokafka/record/legacy_records.py
shargan: True, but that error is already very explicit:
```python
>>> builder = LegacyRecordBatchBuilder(1, None, 16384)
>>> builder.append(0, None, object(), b"value")
  File "<stdin>", line 1, in <module>
  File "/Users/shawnh/repos/aiokafka/aiokafka/record/legacy_records.py", line 304, in append
    key_size = len(key) if key is not None else 0
TypeError: object of type 'object' has no len()
>>> builder.append(0, None, b"key", object())
  File "<stdin>", line 1, in <module>
  File "/Users/shawnh/repos/aiokafka/aiokafka/record/legacy_records.py", line 305, in append
    value_size = len(value) if value is not None else 0
TypeError: object of type 'object' has no len()
```

Python developers should be used to this pattern since type-checking is so expensive in the language. After all, most of the stdlib doesn't bother to check types and just ignores bad values or bubbles up esoteric errors when they're sent around.

aiokafka/record/legacy_records.py
tvoinarovskyi: If I set `max_batch_size=1024 * 1024 * 1024 * 1024` it will get bad. And the name `max_batch_size` does not imply that behaviour. This also differs from the way C extension handles the case, which is not what I want.

aiokafka/record/legacy_records.py
shargan: Three reasons, in order of ascending importance:
1. It removes start_pos from the pack_into calls which makes them a bit clearer.
2. It removes start_pos from the slice on line 382, making it clearer at a glance.
3. It makes the slice on line 382 free (in terms of memory). Slicing bytearrays requires creating new objects while slicing memoryviews does not.

Since creating the memoryview is extremely cheap, I figured the gains were worth it.

tests/record/test_legacy.py
shargan: Lines 201-205:
```python
195     buffer = _make_compressed_batch(magic)
196     buffer = bytearray(buffer)
197     len_offset = 8
198
199     # If the wrapper of compressed messages has a key it will just be ignored.
200     key_offset = 26 if magic else 18
201     new_buffer = (
202         buffer[:key_offset] +
203         b"\x00\x00\x00\x03123" +  # Insert some KEY into wrapper
204         buffer[key_offset + 4:]  # Ignore the 4 byte -1 value for old KEY==None
205     )
```

Memoryviews don't that sort of direct byte addition. Rather than append `.tobytes()` at the end of the buffer slices, I just cast the entire buffer to a `bytearray` at the beginning of the test.

tvoinarovskyi: Just shrink the buffer and return it. No need for the memoryview. I don't want a difference between C and Python in returned types.

aiokafka/record/legacy_records.py
shargan: Is that a common practice?

The C extensions are brand new; I certainly understand wanting to make them behave the same way as the Python code and would be happy to add that diff.

aiokafka/record/legacy_records.py
```diff
@@ -278,155 +285,155 @@ def __repr__(self):
...
+                "I"    # CRC => Int32
+                "b"    # Magic => Int8
+                "b"    # Attributes => Int8
+                "i"    # key length => Int32
+                "%ds"  # key => bytes
+                "i"    # value length => Int32
+                "%ds"  # value => bytes
+                % (key_size, value_size),
+                buf, 0, offset, length, 0, magic, attributes,
+                key_size or -1, key or b"", value_size or -1, value or b"")
```
tvoinarovskyi: There's an error here. The `key_size or -1` will give a false result on `key == b""`. We had to write empty bytes but resulted in null bytes.

shargan: Ahh, great catch. Fixed and added a test for this case.

shargan:
The most recent diff changes some behavior as requested. Does extending the buffer in chunks work for you as a middle ground? We still get most of the performance benefit while giving up a little bit to repeated `len()` calls.

```bash
# fully pre-allocated buffer, no len() calls
$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 616 us +- 43 us
.....................
batch_append_v1: Mean +- std dev: 667 us +- 33 us
$ python bench_records.py simple
1000000 messages in 5.44s (183712.12 msg/s)


# chunk-allocated buffer with len() calls
$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 602 us +- 42 us
.....................
batch_append_v1: Mean +- std dev: 648 us +- 39 us
$ python bench_records.py simple
1000000 messages in 5.64s (177179.21 msg/s)


# no pre-allocation, extended by message size on each call
$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 600 us +- 43 us
.....................
batch_append_v1: Mean +- std dev: 623 us +- 19 us
$ python bench_records.py simple
1000000 messages in 6.34s (157789.73 msg/s)
```

Note that benchmark/record_batch_compose.py shows no positive change from this because it only adds 100 messages per batch. The performance improvement here really becomes apparent as the batches are fully utilized and the cost of extending them surpasses the cost of allocation. To compare, here's the results of that benchmark with `MESSAGES_PER_BATCH=2000`:

```bash
# no pre-allocation, extended by message size on each call
$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 11.3 ms +- 0.5 ms
.....................
batch_append_v1: Mean +- std dev: 13.3 ms +- 0.9 ms


# chunk-allocated buffer with len() calls
$ python benchmark/record_batch_compose.py
.....................
batch_append_v0: Mean +- std dev: 10.7 ms +- 0.3 ms
.....................
batch_append_v1: Mean +- std dev: 12.0 ms +- 0.8 ms
```

aiokafka/record/_legacy_records.pyx
```diff
@@ -429,7 +440,9 @@ cdef class _LegacyRecordBatchBuilderCython:
                     compressed = lz4_encode(bytes(self._buffer))
             else:
                 return 0
-            size = _size_in_bytes(self._magic, key=None, value=compressed)
+            value_size = _size_of_bytearray(compressed)
```
tvoinarovskyi: Could you explain why was the change to C extension made? Does it make a speed difference?

shargan: `_size_in_bytes` was changed to accept lengths rather than objects for the sake of consistency between the Python and C versions. `_size_of_bytearray` was added to avoid reproducing that logic all over the place.

This mattered more in the original version when the signature of `size_in_bytes()` had changed. I'm happy to revert it, if you'd prefer.

tvoinarovskyi: Yes please

tvoinarovskyi:
Hey there, I would like to have this PR merge to release, @shargan If you can/like to apply changes, please say so. If not it's not a problem, will merge like this and make minor tweeks in a followup. Sorry for being too picky with those changes and the ones before, I am very grateful for your work on this)

shargan:
Hey @tvoinarovskyi, sorry for dropping off the radar. Deadlines have prevented me from making progress on this and will likely continue to do so for a couple more weeks. I don't mind the pickiness; I'll be sure to send much smaller PRs in the future to make changes easier to work through. Thanks for working with me on this.

tvoinarovskyi:
@shargan No problem, I did drop off lately myself) Did a followup on this https://github.com/aio-libs/aiokafka/pull/301

Issue #246, opened 2017-10-27
Exception traceback is multiplied by number of futures


Artimi:
When there is raised `RequestTimedOutError` during commiting the `AIOKafkaConnection` is closed and because of that all futures that waits in `AIOKafkaConnection._requests` have set exception. However there is reused single exception for this. And this results in having traceback of this exception strangely multiplied. Take a look at exception traceback that I got:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/messaging_client/mqclient/topic_check.py", line 159, in commit
    await self._consumer.commit(positions)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/consumer.py", line 407, in commit
    yield from self._coordinator.commit_offsets(offsets)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/group_coordinator.py", line 544, in commit_offsets
    node_id, request, group=ConnectionGroup.COORDINATION)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/group_coordinator.py", line 286, in _send_req
    raise err
  File "/usr/local/lib/python3.5/site-packages/aiokafka/group_coordinator.py", line 280, in _send_req
    resp = yield from self._client.send(node_id, request, group=group)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/client.py", line 408, in send
    result = yield from future
  File "/usr/local/lib/python3.5/asyncio/tasks.py", line 392, in wait_for
    return fut.result()
  File "/usr/local/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/local/lib/python3.5/site-packages/messaging_client/mqclient/topic_check.py", line 159, in commit
    await self._consumer.commit(positions)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/consumer.py", line 407, in commit
    yield from self._coordinator.commit_offsets(offsets)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/group_coordinator.py", line 544, in commit_offsets
    node_id, request, group=ConnectionGroup.COORDINATION)
  File "/usr/local/lib/python3.5/site-packages/aiokafka/group_coordinator.py", line 286, in _send_req
    raise err
  File "/usr/local/lib/python3.5/site-packages/aiokafka/group_coordinator.py", line 280, in _send_req
    resp = yield from self._client.send(node_id, request, group=group)
[...]

PR #247, opened 2017-10-27
Instantiate error for each future.set_exception

Artimi:
This fixes https://github.com/aio-libs/aiokafka/issues/246 . I've found that `set_exceptions` is used also in `MessageBatch.failure()` but it is settings an exception that is passed as argument. I would need to change it's interface to pass some function that returns the exception so it can be called for every future in `MessageBatch._msg_future`. Do you agree or do you have another idea?

### DIFF
--- a/aiokafka/producer/message_accumulator.py
+++ b/aiokafka/producer/message_accumulator.py
@@ -1,6 +1,7 @@
 import asyncio
 import collections
 import io
+import copy
 
 from kafka.protocol.types import Int32
 
@@ -165,7 +166,9 @@ class MessageBatch:
         for future, _ in self._msg_futures:
             if future.done():
                 continue
-            future.set_exception(exception)
+            # we need to copy exception so traceback is not multiplied
+            # https://github.com/aio-libs/aiokafka/issues/246
+            future.set_exception(copy.copy(exception))
 
     def wait_deliver(self, timeout=None):
         """Wait until all message from this batch is processed"""
--- a/aiokafka/conn.py
+++ b/aiokafka/conn.py
@@ -174,11 +174,11 @@ class AIOKafkaConnection:
             self._writer = self._reader = None
             self._read_task.cancel()
             self._read_task = None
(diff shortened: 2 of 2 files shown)


tvoinarovskyi:
Sorry for the long wait on this, still checking if there's a more general way of fixing it.

Artimi:
@tvoinarovskyi Have you found something? I was trying to find this bug directly in Python but with no success. This issue is a time bomb for us because it can effectively kill one of our service. What if we use `copy.deepcopy` in `MessageBatch`?

Artimi:
@tvoinarovskyi I added the version with copy. Now it could be ready to merge if you agrees.

aiokafka/producer/message_accumulator.py
```diff
@@ -165,7 +166,9 @@ def failure(self, exception):
         for future, _ in self._msg_futures:
             if future.done():
                 continue
-            future.set_exception(exception)
+            # we need to copy exception so traceback is not multiplied
+            # https://github.com/aio-libs/aiokafka/issues/246
+            future.set_exception(copy.deepcopy(exception))
```
tvoinarovskyi: Seems like copy.copy() is enough to prevent traceback changes. The tracebacks are immutable and linked in a linked list, so copy is enough.
For example:
```python
import copy


def test_raise_many():
    err = ValueError()
    for i in range(100):
        try:
            raise copy.copy(err)
        except Exception:
            pass
    raise err

test_raise_many()
```

tvoinarovskyi:
Sorry, was away for some time. Please use copy and lets merge it.

aiokafka/producer/message_accumulator.py
tvoinarovskyi: See https://gitter.im/aio-libs/Lobby?at=5a0d094871ad3f8736edbd5b reply on gitter also

Artimi:
@tvoinarovskyi Done. Thanks for investigation.

PR #249, opened 2017-10-30
Fix race condition in record fetcher

aerkert:
Concurrent record consuming fails with `InvalidStateError('Exception is not set.')` in record fetcher. This fixes race condition.


tvoinarovskyi:
LGTM, will wait for the build to finish and merge

PR #255, opened 2017-11-08
Make sure Conn._read task exceptions are handled

ask:
Possible fix for #252

The connection starts a background thread using `asyncio.ensure_future` but does not await it
and so exceptions raised by it would be silenced was it not for the Python "Task exception was never retrieved" warning message.

It is still a mystery why the warning involves a `CancelledError` that in theory should have been caught and handled by `_read`'s `except CancelledError: pass` section...  Adding a callback to ensure
any exception is read is anyway a good practice.

I used the same approach as the idle check, and assume the `self.close` will be sufficient to recover.

### DIFF
--- a/aiokafka/conn.py
+++ b/aiokafka/conn.py
@@ -106,10 +106,23 @@ class AIOKafkaConnection:
         self._reader, self._writer, self._protocol = reader, writer, protocol
         # Start reader task.
         self._read_task = ensure_future(self._read(), loop=loop)
+        self._read_task.add_done_callback(self._on_read_task_error)
         # Start idle checker
         if self._max_idle_ms is not None:
             self._idle_handle = self._loop.call_soon(self._idle_check)
         return reader, writer
+
+    def _on_read_task_error(self, read_task):
+        try:
+            read_task.result()
+        except Exception as exc:
+            conn_exc = Errors.ConnectionError(
+                "Connection at {0}:{1} broken".format(self._host, self._port))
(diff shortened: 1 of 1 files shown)


tvoinarovskyi:
This is plain strange. I do agree that we need to log errors for the background task. But how in the world did the Cancelled error go pass that try/except block...

vineet-rh:
@tvoinarovskyi do you think this can be merged into master? Our logs are getting flooded with repeated error logs like this one:
```
[2018-01-24 18:17:47,547: ERROR] Task exception was never retrieved
future: <Task finished coro=<AIOKafkaConnection._read() done, defined at /home/foo/env/lib/python3.6/site-packages/aiokafka/conn.py:194> exception=CancelledError()>
```

tvoinarovskyi:
@vineet-rh I did not really manage to reproduce this, so I'm not sure the fix helps. Will try to resolve it before release.

tvoinarovskyi:
Honestly, I don't see anything bad about this PR, but @vineet-rh how can you get this
```
[2018-01-24 18:17:47,547: ERROR] Task exception was never retrieved
future: <Task finished coro=<AIOKafkaConnection._read() done, defined at /home/foo/env/lib/python3.6/site-packages/aiokafka/conn.py:194> exception=CancelledError()>
```
error I have NO idea =).

vineet-rh:
@TarasLevelUp I only get this in production after things have been running for long.

Issue #267, opened 2017-12-11
Producer cannot recover from broker failure


vineet-rh:
It looks like the AIOKafkaProducer is unable to recover from the broker going down.

I tried with the following code:

```
from aiokafka import AIOKafkaProducer
import asyncio


loop = asyncio.get_event_loop()


producer = AIOKafkaProducer(loop=loop, bootstrap_servers='localhost:9092')


async def main():
    await producer.start()
    i = 0
    while True:
        print(f'Attempting to send')
        await producer.send('fooobarrrrr', b'Hello my friend')
        i += 1
        await asyncio.sleep(1.0)


if __name__ == '__main__':
    loop.run_until_complete(main())
```

Looks like if I take down my locally running broker and bring it back up, the program just goes into a loop with the following log message:
```
Unable to update metadata from []
Unable to update metadata from []
Unable to update metadata from []
Unable to update metadata from []
Unable to update metadata from []
Unable to update metadata from []
```

It looks like the producer is unable to get the cluster broker ids after a broker failure.

vineet-rh:
@tvoinarovskyi is this a known issue? Am I doing something wrong/is there a planned fix?

tvoinarovskyi:
Hey there. Thanks for the bug report. I am a bit stuck with a full refactor of subscription state and a little time on my hands, thus the late response.
To my big surprise, I can reproduce this bug. There was quite some work done on the producer in 0.3.1, so I thought those should go away. For information please post your Broker and aiokafka versions used.
Atm, I can't really dig deep into what's happening, but I am sure we did not have this before. I suspect it's either something introduced in 0.3.1 or a problem with Broker 0.11+.

ask:
I'm reproducing with Kafka 0.11.0.1, aiokafka 0.3.1

Once stopping Kafka, it logs this very fast, like multiple times per second which seems like
a spinloop (well, not really, but a bit faster than necessary).

```
[2018-01-02 15:07:02,394: WARNING]: No broker metadata found in MetadataResponse
[2018-01-02 15:07:02,394: ERROR]: Error fetching metadata for topic posts: <class 'kafka.errors.InvalidReplicationFactorError'>
[2018-01-02 15:07:02,395: ERROR]: Error fetching metadata for topic word-counts-word_counts-changelog: <class 'kafka.errors.InvalidReplicationFactorError'>
[...]

Issue #261, opened 2017-11-23
aiokafka consumer don't reset offset when offset is out of range in specific case.


tfilipczak:
### Version information:
Kafka cluster: 0.10.2.1
aiokafka: 0.3.1
Python: 3.6.1
### Consumer script:
```
import logging
import sys

from aiokafka import AIOKafkaConsumer
import asyncio


logging.basicConfig(format='%(asctime)-15s %(levelname)s %(module)s:%(lineno)s %(name)s: %(message)s',
        level=logging.INFO)
logger = logging.getLogger()


loop = asyncio.get_event_loop()


async def consume():
    consumer = AIOKafkaConsumer(
            'test-topic',
            loop=loop, bootstrap_servers='192.168.1.101:9092',
            group_id='aiokafka-test',
            auto_offset_reset='earliest')
    await consumer.start()
    try:
        async for msg in consumer:
            logger.info("consumed: %s, %s, %s, %s, %s, %s", msg.topic, msg.partition, msg.offset, msg.key, msg.value, msg.timestamp)
    finally:
        await consumer.stop()

loop.run_until_complete(consume())
```

### Bug reproduce flow:
1) Set topic retention in kafka - for example: 5 min.
2) Start producing messages to kafka.
3) Start consumer script - it consume messages from kafka correctly.
4) Stop consumer script (kill) to generate lag in Kafka - messages are still produced.
5) Wait until kafka delete part of data which was not consumed. In that moment our consumer group 'aiokafka-test' has no active members.
```
~$ kafka-consumer-groups.sh --bootstrap-server 192.168.1.101:9092 --group aiokafka-test --describe
Consumer group 'aiokafka-test' has no active members.
TOPIC PARTITION CURRENT-OFFSET LOG-END-OFFSET LAG CONSUMER-ID HOST CLIENT-ID
test-topic 6 3263917 3263942 25 - - -
test-topic 2 3268233 3268253 20 - - -
test-topic 7 3264122 3264139 17 - - -
test-topic 8 3262524 3262553 29 - - -
test-topic 3 3264791 3264810 19 - - -
test-topic 4 3265363 3265386 23 - - -
test-topic 0 3262134 3262160 26 - - -
test-topic 5 3266556 3266575 19 - - -
test-topic 1 3264517 3264539 22 - - -
```
6) Start consumer script - consumer notice that offset is out of range (log), but don't reset it.
In this case consumer don't consume messages in this group_id.
```
[...]

Issue #264, opened 2017-12-04
Consumers unable to recover upon kafka cluster failure


vineet-rh:
I noticed that my application which uses both producers and consumers was unable to recover upon some boxes failing in my kafka cluster. The errors I see in the logs are as follows (with errors closest to the point until which my application was running fine, on top):
- kafka.errors.GroupCoordinatorNotAvailableError: [Error 15] GroupCoordinatorNotAvailableError
- [2017-12-04 10:28:30,621: INFO] OffsetCommit failed for group foo due to a coordinator error (RequestTimedOutError), will find new coordinator and retry
- kafka.errors.RequestTimedOutError: [Error 7] RequestTimedOutError

Not sure why the consumers were unable to recover at this point. What is the ideal way of handling such exceptions within the consumer. It also seems like the consumers don't really raise this exception and only logs them.

vineet-rh:
Verified that the consumer is able to recover on in most cases. 

However there are times when the following error is raised and the consumer crashes:
```
kafka.errors.GroupCoordinatorNotAvailableError: [Error 15] GroupCoordinatorNotAvailableError
```

tvoinarovskyi:
Should be fixed by https://github.com/aio-libs/aiokafka/pull/286.

PR #286, opened 2018-01-17
Refactoring to allow fetcher to always reset offsets properly.

tvoinarovskyi:
This is a major work, that I was postponing for a long time. The issue was that `SubscriptionState` used in ``kafka-python`` is not suited for asyncio, as it is not aware of Futures and operates purely on flags. For example, lets look at this pseudo-code:
```python
async def update_fetch_position(self, partitions):
    pending_resets = []
    for tp in partitions:
        if not self._subscriptions.is_assigned(tp):
            continue
        if self._subscriptions.assignment[tp].awaiting_reset():
            pending_resets.append(ensure_future(self._do_partition_reset(tp)))

    if pending_resets:
        await asyncio.gather(pending_resets)

async def _do_partition_reset(self, tp):
    assert self._subscriptions.is_assigned(tp)

    tp_state = self._subscriptions.assignment[tp]
    reset_strategy = tp_state.reset_strategy

    # ...
```
The code could actually hit the assert in `_do_partition_reset` because created tasks are scheduled and can take some time to start executing. Ie the assignment can change before `_do_partition_reset` is executed.

In `kafka-python` this is never a problem because a thread can never be interrupted. We could emulate the same behaviour by adding locks at different places in the code, leading to all public API to become coroutines. That would break a lot, so instead this PR tries to solve those problems by refactoring `SubscriptionState` to contain some futures, etc.

 some points need addressing:
* [x] Refactor unit tests, that relied on mocking SubscriptionState
* [x] Add more tests for SubscriptionState
* [x] Fix coverage drop, it should be present.
* [x] Document API changes. (commit can raise CommitFailedError now)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -98,5 +98,7 @@ To run tests with a specific version of Kafka (default one is 0.10.1.0) use KAFK
 Test running cheatsheat:
 
  * ``make test FLAGS="-l -x --ff"`` - run until 1 failure, rerun failed tests fitst. Great for cleaning up a lot of errors, say after a big refactor.
- * ``make test FLAGS="-k consumer --no-pull"`` - run only the consumer tests.
+ * ``make test FLAGS="-k consumer"`` - run only the consumer tests.
+ * ``make test FLAGS="-m 'not ssl'"`` - run tests excluding ssl.
+ * ``make test FLAGS="--no-pull"`` - do not try to pull new docker image before test run.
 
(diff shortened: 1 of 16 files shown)


vineet-rh:
@tvoinarovskyi do we have an ETA on when this PR should be merged and released?

tvoinarovskyi:
Doing final adjustments and ready to merge. Wanted some feedback on it before releasing, mainly confirmation that it fixes #261 and #264. Also, I would like to fix #267 before release, started working on it today. I would release is possible to the end of this month, maybe this weekend.

PR #297, opened 2018-01-26
Do not update cluster information if MetadataResponse contains 0 nodes

### LINKED ISSUE #267: Producer cannot recover from broker failure

tvoinarovskyi:
Fixes #267

### DIFF
--- a/aiokafka/client.py
+++ b/aiokafka/client.py
@@ -288,7 +288,7 @@ class AIOKafkaClient:
                 self._md_update_waiter.set_result(None)
             self._md_update_fut = create_future(loop=self._loop)
         # Metadata will be updated in the background by syncronizer
-        return self._md_update_fut
+        return asyncio.shield(self._md_update_fut, loop=self._loop)
 
     @asyncio.coroutine
     def fetch_all_metadata(self):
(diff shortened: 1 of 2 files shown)


vineet-rh:
@TarasLevelUp this is probably just a fix locally when we run with just 1 broker right? I notice a similar issue when we run in production with multiple brokers too.

tvoinarovskyi:
I think this fix should work for both cases. I did not see any additional handling in Java's client

vineet-rh:
Thanks! 👍

PR #337, opened 2018-03-03
Add warning report to pytest and fix many warnings in code

tvoinarovskyi:


aiokafka/consumer/consumer.py
```diff
@@ -998,7 +998,6 @@ def getmany(self, *partitions, timeout_ms=0, max_records=None):
         return records
 
     if PY_35:
-        @asyncio.coroutine
```
tvoinarovskyi: This may break on early Python3.5. Probably a minor version check should be present

tvoinarovskyi: Addressed in followup beeb9f07074dee23d8a8d4184f2b81df0088ca31

Makefile
```diff
@@ -15,10 +15,10 @@ flake:
 	flake8 aiokafka tests $$extra
 
 test: flake
-	py.test -s --no-print-logs --docker-image $(DOCKER_IMAGE) $(FLAGS) tests
+	py.test -s --docker-image $(DOCKER_IMAGE) $(FLAGS) -Wdefault tests
```
asvetlov: FYI: aiohttp uses strict error mode: https://github.com/aio-libs/aiohttp/blob/master/setup.cfg#L14

PR #375, opened 2018-06-05
Test deploy

tvoinarovskyi:


.travis.yml
```diff
@@ -46,9 +50,9 @@ after_success:
 
 deploy:
```
webknjaz: You might want to have a separate stage for deploy, which will only run after all tests.

```yaml
jobs:
  include:
  - stage: Deploy
    before_install: []
    install: skip
    script: skip
    after_success: []
    deploy:  # move it here
```

.travis.yml
```diff
@@ -46,9 +50,9 @@ after_success:
...
   provider: pypi
-  user: andrew.svetlov
+  user: aio-libs-bot
   password:
-    secure: "o1Ws6S+Tc9880TEcDHMLEFbehGmQKvr7BTwoOuVMgZ4WCVt8CS96I2rjFzZIUIxeIwd/+Mgz37XEjbQy1slkOCh2aapCq361HM6L6c/3V6dvx0Kw2mRZzr/GwfgiuR6JersgSYnuu1WnwetQ2YPVhrXWFWDAA3oZs22R9CmWrvo="
+    secure: "TANs/tSxdtMXjFS1Vgp1EuON2ZfvkR+26HOWDzLkITsaSQ7KxbJJ2TBr4f5OI6BTLZmGB+o3pXcphIkAWyfzEBqK8ZGoFDOOz4Lagmc/uA5Foaohw6+2uTDg6kRsCiJl0Io3zCEOwWNn+1ZfLtbkBKpkBXSR/e23YeXq6NuyEAzfEX+KnHexy4gYZv1SqQA06X8c0kJaKCH/ppwE0hW4z7+/jeHnrTxedM8kpLwaGTSsjaQs1tkqR+bJg7pi2iaL8DgB34uVT6lEWBvWd8X8HC3glZvbGYg6gZL5Hb5NdK+pMevlYX2tIA71fILL81Orx1UsbHCH6emnDtQoh8FfZ+FR8IutzaC2T5+dUdNITj9wb+1ALeN1IEAqHnGxve/V62VXsF7HHGV0s3JrdLzA9Ro3k38DKUo1ekinxqQKAULmJfY6qQHX3f61u+QGv6vAGFgrFfDo62ZfB7bixMjX72WkiK+PiDN7jioYsHyobqN8BA9UjzhcMrVk1LjzsPYg1HRRwX8dp+A605Tt3dxlb1hYfVqDoaJmkakUGb2DfSUov8LDI73ufb6/panMHm6LBVeo5MJ8MCVqvj9LZ2xQTubzXunPDY6XBvYgBKKSGkhwShzsI8nIU0oHHDuQAIMuL4v+lN9Pr9zVRx4Drxem05uTXoREFWT6usmPkimQl7I="
   distributions: "sdist"
```
webknjaz: `sdist bdist_wheel`

.travis.yml
```diff
@@ -22,15 +17,22 @@ stages:
 matrix:
   include:
     - stage: *stage_sanity_test
+      sudo: false
+      python: 3.6
       script:
       - make ci-test-unit
 
     - stage: *stage_functional
+      sudo: required
```
webknjaz: Why do you need sudo in the first place?

.travis.yml
```diff
@@ -22,15 +17,22 @@ stages:
...
+      python: 3.6
+      services:
+      - docker
       env: KAFKA_VERSION=0.11.0.0 SCALA_VERSION=2.12 PYTHONASYNCIODEBUG=1
       script:
       - make ci-test-all
 
     - stage: *stage_deploy
```
webknjaz: Please also add:
```yaml
if: tag IS present
```

.travis.yml
webknjaz: Oh, you already have in the stages ordering section, so nevermind.

.travis.yml
```diff
@@ -1,56 +1,67 @@
...
-    - /^v\d+\.\d+\.\d+$/
+  - master
+  - v0.2
+  - "/^v\\d+\\.\\d+\\.\\d+$/"
+  - "/^v\\d+\\.\\d+\\.\\d+\\.dev\\d*$/"
+
+stages:
+- &stage_sanity_test Unit tests on container workers
+- &stage_functional Run main functional tests
+- &stage_deploy Optionally run deploy
```
webknjaz: You need `name: ` here, otherwise it is an invalid yaml

.travis.yml
tvoinarovskyi: Can it run docker without sudo?

webknjaz: It looks like Travis does this implicitly:

> Travis CI automatically routes builds to run on our Trusty sudo-enabled infrastructure when services: docker is configured. We do not currently support use of Docker on OS X.

tvoinarovskyi: Yea, I can't plug kafka to a sudo: false build either.

.travis.yml
```diff
@@ -1,56 +1,69 @@
-language: python
+version: "= 0"
```
webknjaz: This is not needed anymore. But use `conditions: v1` instead.

tvoinarovskyi:
@webknjaz Thanks for the help on this. Could you take one final look at it?

PR #392, opened 2018-07-05
remove py3.3 support

### LINKED ISSUE #390: remove python 3.3 support as it reached EoL
https://devguide.python.org/#status-of-python-branches
Python 3.3 reached End of Life in 2017-09-29 and no longer is supported in any way by mainstream. No need to bother with it.

rooterkyberian:
resolves #390

### DIFF
--- a/tests/_testutil.py
+++ b/tests/_testutil.py
@@ -218,13 +218,6 @@ class KafkaIntegrationTestCase(unittest.TestCase):
             keyfile=str(self.ssl_folder / "cl_client.key"),
             password="abcdefgh")
 
-    def assertLogs(self, *args, **kw):  # noqa
-        if sys.version_info >= (3, 4, 0):
-            return super().assertLogs(*args, **kw)
-        else:
-            # On Python3.3 just do no-op for now
-            return self._assert_logs_noop()
-
     @contextmanager
     def _assert_logs_noop(self):
         yield
--- a/tests/_testutil.py
+++ b/tests/_testutil.py
@@ -5,7 +5,6 @@ import time
 import unittest
 import pytest
 import operator
-import sys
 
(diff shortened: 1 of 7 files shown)


tests/_testutil.py
```diff
@@ -218,13 +218,6 @@ def create_ssl_context(self):
...
 
-    def assertLogs(self, *args, **kw):  # noqa
-        if sys.version_info >= (3, 4, 0):
-            return super().assertLogs(*args, **kw)
-        else:
-            # On Python3.3 just do no-op for now
-            return self._assert_logs_noop()
-
     @contextmanager
     def _assert_logs_noop(self):
```
webknjaz: Is this used somewhere else?

tests/test_helpers.py
```diff
@@ -21,37 +21,9 @@ def _check_ssl_dir(self):
         self.assertTrue(keyfile.exists(), str(keyfile))
         return cafile, certfile, keyfile
 
-    @pytest.mark.skipif(sys.version_info >= (3, 4),
```
webknjaz: How does this relate to 3.3?

tests/_testutil.py
rooterkyberian: nope, good catch

tests/test_helpers.py
rooterkyberian: description was wrong, this test was 3.3 specific, as indicated by `skipif(sys.version_info >= (3, 4)`

rooterkyberian:
Coverage dropped by one line, not sure what exactly do with it, as it is a support for old 3.4 version (pre 3.4.4) missing `ensure_future` https://docs.python.org/3/library/asyncio-task.html#asyncio.ensure_future

Issue #365, opened 2018-05-10
how to consume only one partition of a topic


hongweishang:
import asyncio
import aiokafka
from aiokafka import AIOKafkaConsumer
from aiokafka import TopicPartition

tp = TopicPartition('my_topic', 0)
loop = asyncio.get_event_loop()
consumer = AIOKafkaConsumer(loop=loop, bootstrap_servers='localhost:9091')
consumer.assign([tp])
async def f():
        async for msg in consumer:
            print(msg)
asyncio.get_event_loop().run_until_complete(f())

when i run this python file,  traceback:
raceback (most recent call last):
  File "t.py", line 17, in <module>
    asyncio.get_event_loop().run_until_complete(f())
  File "/usr/local/python3.6.5/lib/python3.6/asyncio/base_events.py", line 468, in run_until_complete
    return future.result()
  File "t.py", line 13, in f
    async for msg in consumer:
  File "/xxx/venv/lib/python3.6/site-packages/aiokafka/consumer/consumer.py", line 1018, in __anext__
    return (yield from self.getone())
  File "/xxx/venv/lib/python3.6/site-packages/aiokafka/consumer/consumer.py", line 949, in getone
    msg = yield from self._fetcher.next_record(partitions)
AttributeError: 'NoneType' object has no attribute 'next_record'.

why?

tvoinarovskyi:
You have to run `await consumer.start()`

hongweishang:
@tvoinarovskyi thanks，but it does not work.
I add "start" sentence, and AssertionError is raised.

New file is shown as below:
`import asyncio
import aiokafka
from aiokafka import AIOKafkaConsumer
from aiokafka import TopicPartition

tp = TopicPartition('my_topic', 0)
loop = asyncio.get_event_loop()
consumer = AIOKafkaConsumer(loop=loop, bootstrap_servers=['localhost:9091', 'localhost:9092'])
consumer.assign([tp])
async def f():
       await consumer.start()
       async for msg in consumer:
       print(msg)
asyncio.get_event_loop().run_until_complete(f())`


traceback is shown as below:
Traceback (most recent call last):
  File "t.py", line 20, in <module>
    asyncio.get_event_loop().run_until_complete(f())
  File "/usr/local/python3.6.5/lib/python3.6/asyncio/base_events.py", line 468, in run_until_complete
    return future.result()
  File "t.py", line 14, in f
    await consumer.start()
[...]

PR #395, opened 2018-07-07
SubscriptionState: Update assert on assign to include USER_ASSIGNED. …

### LINKED ISSUE #394: Manual partition consuming is broken
It doesn't look like you can use Manual Partition Consuming without hitting an assert:

```
Application error
  File "/Users/nimi/.pyenv/versions/3.6.3/lib/python3.6/site-packages/grpclib/server.py", line 312, in request_handler
    await method.func(stream)
  File "./client-python/main.py", line 85, in Start
    await self.consumer.start()
  File "/Users/nimi/.pyenv/versions/3.6.3/lib/python3.6/site-packages/aiokafka/consumer/consumer.py", line 270, in start
    self._coordinator.assign_all_partitions(check_unknown=True)
  File "/Users/nimi/.pyenv/versions/3.6.3/lib/python3.6/site-packages/aiokafka/consumer/group_coordinator.py", line 103, in assign_all_partitions
    self._subscription.assign_from_subscribed(partitions)
  File "/Users/nimi/.pyenv/versions/3.6.3/lib/python3.6/site-packages/aiokafka/consumer/subscription_state.py", line 202, in assign_from_subscribed
    SubscriptionType.AUTO_PATTERN, SubscriptionType.AUTO_TOPICS]
```

Basically `start` triggers an asset that the SubscriptionType is either `SubscriptionType.AUTO_PATTERN, SubscriptionType.AUTO_TOPICS`

## Reproduction

```python
consumer = AIOKafkaConsumer(
            loop=self.loop,
            bootstrap_servers=request.kafka_brokers,
            enable_auto_commit=False,
            group_id=None,
        )
consumer.assign(TopicPartition('foo', 0))
await consumer.start()
```

## Cause

When you call `consumer.assign` that sets changes the `consumer`'s [subscription to ManualSubscription](https://github.com/aio-libs/aiokafka/blob/9659b380c9a28923fe3ae3137784dff411686dfe/aiokafka/consumer/subscription_state.py#L169) with a subscription type of `SubscriptionType.USER_ASSIGNED`.

```python
def assign_from_user(self, partitions: Set[TopicPartition]):
        """ Manually assign partitions. After this call automatic assignment
        will be impossible and will raise an ``IllegalStateError``.
        Caller: Consumer.
        Affects: SubscriptionState.subscription
        """
        self._set_subscription_type(SubscriptionType.USER_ASSIGNED)

        self._change_subscription(
            ManualSubscription(partitions, loop=self._loop))
        self._notify_assignment_waiters()
```

Next, when you call `consumer.start` we then check if `consumer.subscription` is not `None`, and then call `NoGroupCoordinator .assign_all_partitions`.

```python
self._coordinator = NoGroupCoordinator(
    self._client, self._subscription, loop=self._loop,
    exclude_internal_topics=self._exclude_internal_topics)

# If we passed `topics` to constructor.
if self._subscription.subscription is not None:
    yield from self._client.force_metadata_update()
    self._coordinator.assign_all_partitions(check_unknown=True)
```

Which then calls the base class `Subscription` `assign_from_subscribed`:

```python
def assign_from_subscribed(self, assignment: Set[TopicPartition]):
        """ Set assignment if automatic assignment is used.
        Caller: Coordinator
        Affects: SubscriptionState.subscription.assignment
        """
        assert self._subscription_type in [
            SubscriptionType.AUTO_PATTERN, SubscriptionType.AUTO_TOPICS]

        self._subscription._assign(assignment)
        self._notify_assignment_waiters()
```

Which calls the assert. My hunch is that the assert should also include `SubscriptionType.USER_ASSIGNED`, however does adding that to the assert break any invariants required by `_assign` and `_notify_assignment_waiters`?

nemosupremo:
…Fixes #394, #365

### DIFF
--- a/aiokafka/consumer/group_coordinator.py
+++ b/aiokafka/consumer/group_coordinator.py
@@ -106,8 +106,9 @@ class NoGroupCoordinator(BaseCoordinator):
         assignment = self._subscription.subscription.assignment
         for tp in partitions:
             tp_state = assignment.state_value(tp)
-            tp_state.reset_committed(
-                OffsetAndMetadata(UNKNOWN_OFFSET, ""))
+            if tp_state:
+                tp_state.reset_committed(
+                    OffsetAndMetadata(UNKNOWN_OFFSET, ""))
 
     @property
     def _group_subscription(self):
--- a/aiokafka/consumer/subscription_state.py
+++ b/aiokafka/consumer/subscription_state.py
@@ -205,7 +205,8 @@ class SubscriptionState:
         Affects: SubscriptionState.subscription.assignment
         """
         assert self._subscription_type in [
-            SubscriptionType.AUTO_PATTERN, SubscriptionType.AUTO_TOPICS]
+            SubscriptionType.AUTO_PATTERN, SubscriptionType.AUTO_TOPICS,
+            SubscriptionType.USER_ASSIGNED]
 
         self._subscription._assign(assignment)
         self._notify_assignment_waiters()


tvoinarovskyi:
Hi, thanks for the work on this! It may break some corner cases to remove that assert, so closing in favour of #424

PR #405, opened 2018-07-31
Add tests against Kafka 1.0.1

castarco:
Just to improve the confidence that we as developers can have on AIOKafka, it would be nice to see how it works against Kafka 1.0.1.


webknjaz:
Tests fail

castarco:
Yeah, we should push a new Docker image to the DockerHub registry for Kafka 1.0.1. But strangely TravisCI is not complaining about the missing Docker image.

webknjaz:
Probably pytest suppresses that or somehow picks odd image. Can you reproduce this locally?

castarco:
I'll try to check it this evening (Berlin time)

tvoinarovskyi:
@castarco Thanks for the PR, I will need to build and update an image for Kafka 1.0.1. It requires custom docker image to build.

tvoinarovskyi:
I've fully updated travis.yml, so will fix this along with #413. Will be 1.0.2 Kafka, hope you don't mind

PR #434, opened 2018-09-27
Transactional producer

tvoinarovskyi:
Checklist:

* [ ] Write basic test cases for transactional producer behaviour:
  * [ ] Test 2 producers fencing one another by using the same transactional_id
  * [ ] Test that a restarted producer will get the same PID if it's using the same transactional_id
  * [ ] Ensure we raise OutOfOrderSequence if we lose some messages. This can be done by bumping the sequence manually.
  * [ ] Ensure the transaction will materialise only after we commit in READ_COMMITTED and straight away if READ_UNCOMMITTED
  * [ ] Ensure the consumer will not yield control messages to the user
* [ ] Add support for Transaction Coordinator in InitPID request. The InitProducerIdRequest should be only routed to Coordination node. We will need a mechanism similar to Consumer's coordinator lookup.
* [ ] Add **begin_transaction** API that will create a new transaction and mark all subsequent sends with it. 
* [ ] After **begin_transaction** we need to add each partition to the transaction as new sends are done. Use AddPartitionsToTxn.
* [ ] Add **send_offsets_to_transaction** API that will be able to add a new commit to the current transaction. Calling this method not within a transaction is prohibited. There can be only 1 transaction per Producer at a time.
* [ ] Add **end_transaction** API that will flush all pending messages in the Producer and commit transaction.
* [ ] Add **init_transactions** API that will abort previous transactions on the same transactional_id.
* [ ] Add as ``async with`` API simplification. **init_transactions** will be called at first entry, **begin_transaction** on `__enter__` and  **end_transction** on exit. Commit should still remain explicit. 
```python
producer = AIOKafkaProducer(..., transactional_id="my-id")
producer.start()
async with producer.transaction() as t:
    for i in range(10):
        producer.send("topic", b"Super Data")
    producer.send_offsets_to_transaction({TopicPartition("topic-0", 0): 1})
```
* [ ] Add last_stable_offset API to consumer

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -81,19 +81,21 @@ Running tests
 
 Docker is required to run tests. See https://docs.docker.com/engine/installation for installation notes. Also note, that `lz4` compression libraries for python will require `python-dev` package,
 or python source header files for compilation on Linux.
+NOTE: You will also need a valid java installation. It's required for the ``keytool`` utility, used to
+generate ssh keys for some tests.
 
 Setting up tests requirements (assuming you're within virtualenv on ubuntu 14.04+)::
 
     sudo apt-get install -y libsnappy-dev
     make setup
 
-Running tests::
+Running tests with coverage::
 
     make cov
 
-To run tests with a specific version of Kafka (default one is 0.10.1.0) use KAFKA_VERSION variable::
+To run tests with a specific version of Kafka (default one is 1.0.2) use KAFKA_VERSION variable::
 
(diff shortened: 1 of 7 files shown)


tvoinarovskyi:
Fixes #441

tvoinarovskyi:
The only issue remaining is to fix coverage and find a way to avoid catching the commit point in consumer, as it may be committed from producer now.

sibiryakov:
well done!

PR #458, opened 2018-12-27
Fix speed degradation introduced with wait_for_reponse_or_error on send()

tvoinarovskyi:


tvoinarovskyi:
Wrapping the future on send every time was a bit too expensive

PR #386, opened 2018-06-27
Feat: Add SASL PLAIN authentication support

cyrbil:
https://github.com/aio-libs/aiokafka/issues/84

PR #464, opened 2018-12-30
Add sasl plain support to testing environment

### LINKED ISSUE #84: Add SASL Plain support
kafka-python already has support for that, so it shouldn't be that hard.

tvoinarovskyi:
Mostly based on #386 from @cyrbil. Fixes #84


tvoinarovskyi:
Yea, so this needs more work in Consumer. Currently, the code does not propagate critical errors to the user too well. Need to write handling for all authorization errors to be passed correctly. Right now it crashes the `_coordination_routine`.

tvoinarovskyi:
Should be fully functional at this point. Will go over another one of review and coverage cleanup and will merge.

PR #469, opened 2019-01-16
SASL GSSAPI/Kerberos support

jsurloppe:
Hi,

I've take the opportunity that's you're currently working on SASL_PLAIN to add GSSAPI to your branch.

I've tested my branch on confluent cp-docker-images/examples/kafka-cluster-sasl (PLAINTEXT) and our production cluster (SSL), still doing some tests but i'm not encountering any problem for now.

### DIFF
--- a/aiokafka/conn.py
+++ b/aiokafka/conn.py
@@ -261,16 +261,30 @@ class AIOKafkaConnection:
             self.log.warning(
                 'Sending username and password in the clear')
 
-        authenticator = self.authenticator_gssapi() \
+        q_req = asyncio.Queue(maxsize=1)
+        q_resp = asyncio.Queue(maxsize=1)
+
+        authenticator = self.authenticator_gssapi(q_req, q_resp) \
             if self._sasl_mechanism == 'GSSAPI' \
-            else self.authenticator_plain()
+            else self.authenticator_plain(q_req, q_resp)
+
+        task = self._loop.create_task(authenticator)
+
+        def exception_handler(fut):
+            exc = fut.exception()
+            if exc:
+                q_resp.put_nowait((exc, False))
+
+        task.add_done_callback(exception_handler)
 
         auth_bytes = None
-        while True:
-            try:
-                payload, expect_response = authenticator.send(auth_bytes)
-            except StopIteration:
-                break
+        expect_response = True
+
+        while not task.done():
+            q_req.put_nowait(auth_bytes)
+            payload, expect_response = yield from q_resp.get()
+            if isinstance(payload, Exception):
+                raise payload
 
             if req_klass.API_VERSION == 0:
                 auth_bytes = yield from self._send_sasl_token(payload,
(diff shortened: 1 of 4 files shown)


aiokafka/conn.py
```diff
@@ -18,10 +19,18 @@
 import aiokafka.errors as Errors
 from aiokafka.util import ensure_future, create_future, PY_341, PY_36
 
+from kafka.protocol.types import Int8
```
tvoinarovskyi: Please just use `struct` package directly

aiokafka/conn.py
```diff
@@ -262,8 +289,39 @@ def _do_sasl_handshake(self):
...
+        server_token = None
+        while not client_ctx.complete:
+            client_token = client_ctx.step(server_token)
+            if client_token:
+                server_token = yield client_token, True
+            else:
+                server_token = yield b'', True
+        msg = client_ctx.unwrap(server_token).message
+
+        iomsg = io.BytesIO(msg[0:1])
```
tvoinarovskyi: Int8 is just 1 byte, right? msg[0] will give you the value already. This is full Python3, so no need for this wrapping.

jsurloppe:
Totally agree that we should have a docker test env with all the stack for testing againt next versions.

For now the thing the only thing that annoy me is that gssapi client_ctx.step() trigger indirect blocking IO (krb libs exchanging with KDC), it's only used for ticket exchange so i'm not sure if it's worth an executor...

tvoinarovskyi:
It is worth an executor. It's a network call, right? So if any kind of network issue happens it means we will end up with a stop in our application. That is no good.

aiokafka/conn.py
```diff
@@ -216,7 +233,11 @@ def _do_version_lookup(self):
 
     @asyncio.coroutine
     def _do_sasl_handshake(self):
-        req_klass = self._version_info.pick_best(SaslHandShakeRequest)
+        if self._sasl_mechanism == "GSSAPI":
+            # v1 is unsupported, trigger a infinite auth loop
```
tvoinarovskyi: I feel like this is not true. The v1 is supported, we just need to figure out how it's done. It works with Java client )

tvoinarovskyi:
@jsurloppe As I see this is based on kafka-python implementation. I know the author spent many hours there setting that up, so I feel like we can go with this implementation, it's still better than none. So we can delegate automatic testing of this feature, as for executor and v1 support I can totally merge without them if you feel like those would be hard to implement.

aiokafka/conn.py
jsurloppe: dunno, both kafka-python and rdkafka use v0

jsurloppe:
I tried different approach but the executor is pretty tricky, futures don't like to be passed between loops and generated based coroutine won't allow async generator.

We might be able to have something cleaner if we move to new syntax and async gen 3.6+

As for v1, not sure to find the time to investigate at the moment, tbh.

tvoinarovskyi:
@jsurloppe Cool, thanks. Will merge this and do a followup after if any fixes needed. I have no idea why travis did not build for a fork thou...

aiokafka/conn.py
tvoinarovskyi: Lol, version v1 did not work for such a dumb reason. Basically a few of my buggy lines of code where req_klass is used in 2 places and next request just always ended up with API_VERSION=0 =). It works for both versions, kafka-python and rdkafka just did not pick this up yet.

PR #475, opened 2019-01-27
Add Kerberos GSSAPI support

tvoinarovskyi:


tvoinarovskyi:
@jsurloppe Could you take a final look at it and check if it works on your setup? Mostly it should be finished, I probably failed to upload 0.11.0.3 broker, so travis broke. If we're ok with it, will try to cut a release this week.

jsurloppe:
Just run some test on our cluster, everything works perfectly!

PR #478, opened 2019-02-13
Use MetadataRequest_V1 if possible during client bootstrap

### LINKED ISSUE #440: Producer sends metadata requests about all topics on kafka broker
We've recently started using ACLs for our Kafka cluster, and we're seeing lots of spam in our kafka logs about aiokafka trying to get metadata for all topics, and being denied because of these ACLs.

Everything seems to be working fine, it just seems unnecessary to try and get metadata for stuff the producer can't access anyway.

I think I've found a workaround for now, which is to set the topic the producer is supposed to use with `producer.client.set_topics([<topic>])` before starting the producer with `start()`, but I'm not 100% certain this will not cause any side effects.

Is this a safe thing to do in order to get rid of all the ACL spam in our kafka logs?

Update: seems related to #287

ulrikjohansson:
This is a naive fix for the problem of client bootstrap trying to fetch all topic metadata.
It only changes current behaviour if a version is manually supplied to the client.

v1 of the protocol changes the semantics of [] to mean "don't fetch any topic metadata at all", v0 [] means "get all topic metadata".

Is this good enough as a temporary solution?

Fixes #440

### DIFF
--- a/aiokafka/client.py
+++ b/aiokafka/client.py
@@ -166,8 +166,12 @@ class AIOKafkaClient:
     @asyncio.coroutine
     def bootstrap(self):
         """Try to to bootstrap initial cluster metadata"""
-        # using request v0 for bootstap (bcs api version is not detected yet)
-        metadata_request = MetadataRequest[0]([])
+        # using request v0 for bootstap if not sure v1 is available
+        if self.api_version == "auto" or self.api_version < (0, 10):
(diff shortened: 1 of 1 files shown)


tvoinarovskyi:
Thanks, it was on my list of things for the next release! The fix seems good to me, don't see any downside. The only place where this might have an impact is the pattern subscription, as it requires all topics to be fetched. But that's a different issue

PR #484, opened 2019-03-03
Kerberos test

jsurloppe:
This PR add a GSSAPI unit test and the required docker setup.

I also added some optional param on producer that i've forget last time.


jsurloppe:
Thanks, yeah the whole kerberos things is pretty obscure to setup from scratch when you don't have some real deployment you can refer to because you know it's works. (mostly scratched my head on a dns issue in/out of the container)

PR #504, opened 2019-05-07
fix event_waiter reference before assignment

romantolkachyov:
## What do these changes do?

Fix an error with not defined variable.

Users tests may fail in case of fast consumer shutdown.


tvoinarovskyi:
Thanks!

PR #530, opened 2019-07-11
Message Accumulator: add_message being recursive eventually overflows

ask:
MessageAccuulator.add_message is recursive. If left this way it may lead to stack overflow, which happened to us in production.


tvoinarovskyi:
The original idea was that the 2nd recursive call will always insert the message, but after some changes it seems like it's not the case anymore, there are race conditions. I don't mind having a loop there, thanks!

PR #531, opened 2019-07-11
Consumer.seek_to_committed now returns mapping of committed offsets

ask:
This method is fetching the currently committed offsets, so having it return them saves
us from doing it again.


tvoinarovskyi:
Makes sense. Don't mind the coverage, it does not want to load it from travis.com for some reason. Will try to fix it once I have a minute.

PR #539, opened 2019-08-10
Fix issue #536 

tvoinarovskyi:
For race condition between metadata update and subscription change

### DIFF
--- a/aiokafka/consumer/consumer.py
+++ b/aiokafka/consumer/consumer.py
@@ -1046,6 +1046,14 @@ class AIOKafkaConsumer(object):
             self._subscription.subscribe(
                 topics=topics, listener=listener)
             self._client.set_topics(self._subscription.subscription.topics)
+            if self._group_id is None:
+                # We have reset the assignment, but client.set_topics will not always
+                # do a metadata update. We force it to do it even if metadata did not
+                # change. This will trigger a reassignment on NoGroupCoordinator, but
+                # only if snapshot did not change, thus we reset it too.
+                self._client.force_metadata_update()
+                if self._coordinator is not None:
+                    self._coordinator._metadata_snapshot = {}
             log.info("Subscribed to topic(s): %s", topics)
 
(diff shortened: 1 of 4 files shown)


rjemanuele:
This patch works for a few iterations and then several exceptions are raised.  Test code and logs below:

consumer.py
```
import asyncio
import random
import logging

consumer = None


async def consume():
    global consumer

    consumer = AIOKafkaConsumer(
        loop=loop, bootstrap_servers='kafka.bus:9092', metadata_max_age_ms=5000
    )
    # Get cluster layout and topic/partition allocation
    await consumer.start()
    try:
        async for msg in consumer:
            logging.info("consumed: %s" % msg.value)
    finally:
        await consumer.stop()


async def topic_updater():
    while True:
        i = random.randint(1, 3)
        topic = "my_topic{}".format(i)
        if consumer is not None:
            logging.info('subscribing: %s' % topic)
            consumer.subscribe([topic, 'mytopic-static'])
        await asyncio.sleep(10)

loop.set_debug(True)
my_format = "%(asctime)s - %(levelname)s : %(name)s : %(message)s"
logging.basicConfig(level=logging.DEBUG, format=my_format)
loop.run_until_complete(asyncio.gather(consume(), topic_updater()))
```

[log.txt](https://github.com/aio-libs/aiokafka/files/3489497/log.txt)

tvoinarovskyi:
@rjemanuele Great, this seems to be a separate issue. It does update subscription successfully but failed later on. Have no idea how to fix this one yet.

rjemanuele:
Hi @tvoinarovskyi, Any update on this?

tvoinarovskyi:
@rjemanuele Sorry, not really, hard to find time atm. Thanks for pinging, will try to get some time this week!

tvoinarovskyi:
@rjemanuele Reworked the patch, turned out dirty, but should cover your case fully now. Sorry for taking forever 🙇

PR #548, opened 2019-08-24
Fix batch producer docs example typo

hellysmile:
## What do these changes do?

Fix batch producer docs example typo

No

## Related issue number

No

## Checklist

=\

### DIFF
--- a/docs/examples/batch_produce.rst
+++ b/docs/examples/batch_produce.rst
@@ -27,7 +27,7 @@ Producer
         while i < num:
             msg = ("Test message %d" % i).encode("utf-8")
             metadata = batch.append(key=None, value=msg, timestamp=None)
-            if metadata is None:
+            if metadata is not None:
                 partitions = await producer.partitions_for(topic)
                 partition = random.choice(tuple(partitions))
                 await producer.send_batch(batch, topic, partition=partition)


tvoinarovskyi:
@hellysmile Why should that be "is not"? It appends to the batch until it's full and only sends the batch in that case. Full is notified by "metadata is None"

hellysmile:
@tvoinarovskyi my local example works exactly vise versa

I'll try to prepare full example to reproduce

hellysmile:
@tvoinarovskyi thanks a lot, I've found my issue

From example is not so clear that after `continue` there is appending one more time exact same message to the newly created batch

Changing to `is not None` was just pushing each messages alone in individual batch in my case

If cycle is `for` code becomes little bit more complex

I've ended up with something like 

```python
        batch = kafka_producer.create_batch()

        for i in range(1000):
            msg = {
                'key': None,
                'value': str(i).encode('utf-8'),
                'timestamp': None,
            }

            while True:
                metadata = batch.append(**message)

                if metadata is None:
                    await kafka_producer.send_batch(
                        batch=batch,
                        topic=topic,
                        partition=await get_partition(topic),
                    )

                    batch = kafka_producer.create_batch()

                    continue

                break

        await kafka_producer.send_batch(
            batch=batch,
            topic=topic,
            partition=await get_partition(topic),
        )
```

What about to extend docs regarding 

> Full is notified by "metadata is None"

And we should append our message(exact same message/one more time) to the new batch, cuz I expected that batch became full after adding my message and it contains my message

PR #552, opened 2019-09-05
Add connection open/close callbacks

oded-zahavi:
## What do these changes do?

Add new callbacks for connection opened/close for consumer and producer

User may add callbacks if they want to be notified for changes in connection status (and take action accordingly)

## Related issue number

No

## Checklist

- [x ] I think the code is well written
- [ x] Unit tests for the changes exist
- [ x] Documentation reflects the changes
- [ x] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 CHANGES
 -------
 
+552.feature
+^^^^^^^^^^^
+
+Add new callbacks for connection opened/close for consumer and producer 
+
 523.feature
 ^^^^^^^^^^^
 
--- a/aiokafka/client.py
+++ b/aiokafka/client.py
@@ -137,6 +137,8 @@ class AIOKafkaClient:
         self._md_update_fut = None
         self._md_update_waiter = create_future(loop=self._loop)
         self._get_conn_lock = asyncio.Lock(loop=loop)
+        self._on_connection_closed_callback = None
+        self._on_connection_opened_callback = None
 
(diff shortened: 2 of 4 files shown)


tvoinarovskyi:
Good day, thanks for looking to contribute. Can you explain the use case, where you need to have this callback? Maybe there's a more generic solution for those cases.

oded-zahavi:
Hi @tvoinarovskyi the purpose is to provide the calling application with connection status change indication, in order to allow it to take more actions if needed (such as, but not limited to, event log for instance). I have look thoroughly into the code and could not figure out an elegant way to do that. Callbacks, as suggested, would be a very nice and elegant way to do that. I've tried to write tests for it but figured it would be hard to simulate a change in connection status, and mocking it would be simply not interesting enough. It would be grate to have your opinion!

tvoinarovskyi:
I would prefer some reusable generic interface for similar cases. In java client they have this concept of Interceptors. I would consider something similar instead of adding callbacks for all cases.

oded-zahavi:
There are multiple ways to achieve that. Could you elaborate?

oded-zahavi:
@tvoinarovskyi

oded-zahavi:
@tvoinarovskyi ⬆️

tvoinarovskyi:
Sorry 😢 
I had something like https://aiohttp.readthedocs.io/en/stable/web_advanced.html#signals in mind. What do you think?

PR #593, opened 2020-03-07
kafka-python 2.0 support

yumendy:
## What do these changes do?

kafka-python 2.0 support

change ConnectionError to KafkaConnectionError.

## Related issue number

590.feature


tvoinarovskyi:
Thanks!

PR #595, opened 2020-03-15
Add some missing test, that was in a different PR for some reason.

tvoinarovskyi:
Revert the kafka broker start timeout in tests to 60s


tvoinarovskyi:
ACL for topic creation work very strangely in Kafka. I could not get the KIP-277 Fine grained ACL for topic creation to work at all. Maybe will revisit this issue at a later day.

Issue #599, opened 2020-04-03
aiokafka bdist_rpm command fails


gabriel-tincu:
**Describe the bug**
running `python setup.py bdist_rpm` will fail during the collect debug information step,
namely when `/usr/lib/rpm/find-debuginfo.sh` will fail to retrieve any files from the generated `debugsources.list` file

**Expected behaviour**
I expect the rpm build step to succeed , regardless of missing debug info files. My fix basically ignores the generate debug info step,  which seems to be the source of the issue. While i am fully aware that this is not as much a fix as it is a workaround, the effort involved to actually figure out the internals of rpmbuild would be large and the generated debug packages is for all intended purposes useless

**Environment (please complete the following information):**
 - aiokafka version 0.5.2
 - kafka-python version 1.4.3
 - OS: Fedora 31
**Reproducible example**
```
python setup.py bdist_rpm
```

gabriel-tincu:
would be nice if someone had a look at the PR open for this

tvoinarovskyi:
@gabriel-tincu Sorry for the long wait, the fix seems nice, thanks! Are there any other issues with bdist? I never build an rpm package for it...

gabriel-tincu:
> @gabriel-tincu Sorry for the long wait, the fix seems nice, thanks! Are there any other issues with bdist? I never build an rpm package for it...

Not really... tbh im very much a beginner when it comes to RPM's but in this case it was either this, or adopting a custom solution downstream... This is still a workaround rather than a fix unfortunately :(. Thanks for looking at it @tvoinarovskyi

PR #600, opened 2020-04-03
  Ignore debug package generation on bdist_rpm command

gabriel-tincu:
because the command fails on the debug file information gathering step

https://github.com/aio-libs/aiokafka/issues/599


tvoinarovskyi:
Seems cool, thanks!

Issue #579, opened 2020-02-04
i got same error, i use aiokafka==0.5.2 kafka-python==1.4.6


Achillesxu:
i got same error, i use aiokafka==0.5.2 kafka-python==1.4.6

UnknownMemberIdError

`data = await in_handle.getmany(timeout_ms=1000, max_records=settings.KAFKA_MAX_RECORD)
  File "/root/.pyenv/versions/3.7.6/envs/python3.7.6/lib/python3.7/site-packages/aiokafka/consumer/consumer.py", line 1147, in getmany
    self._coordinator.check_errors()
  File "/root/.pyenv/versions/3.7.6/envs/python3.7.6/lib/python3.7/site-packages/aiokafka/consumer/group_coordinator.py", line 302, in check_errors`

_Originally posted by @Achillesxu in https://github.com/aio-libs/aiokafka/issues/575#issuecomment-581785688_

tvoinarovskyi:
Should be fixed in aiokafka==0.6.0. Please open a new issue or reopen this one if the problem persists

PR #611, opened 2020-05-11
UnknownMemberId was actually raised to the user instead of retrying on commit

tvoinarovskyi:


tvoinarovskyi:
Fixes #579 and #611

PR #619, opened 2020-05-21
Set language_level to version 3.

jameshilliard:
Fixes:
```
FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release!
```


tvoinarovskyi:
I did not even know you can set the language level =)

Issue #618, opened 2020-05-19
Support SASL/OAuthBearer Authentication


oulydna:
Kafka supports SASL/OAuthBearer as an authentication method [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876).

We would really like to have this feature in this library to avoid a separate auth flow for Kafka from other pieces in our environment. So I am creating this issue to get the ball rolling (unless it's already being worked on or planned).

oulydna:
Do we have any plan or not for this?

tvoinarovskyi:
@oulydna Hi there, for now I don't have any plans to work actively on adding it, but will happily review or assist if someone is willing to take a stab at it)

oulydna:
Thanks @tvoinarovskyi for confirming that. Having a PR coming your way.

akvadrako:
Isn't this fixed? I see it was merged years ago.

PR #630, opened 2020-06-10
Added SASL/OAuthBearer

oulydna:
### Changes

Fixes issue #618

Added `OAUTHBEARER` as a new `sasl_mechanism`. The implementation is mainly based on [this PR in kafka-python](https://github.com/dpkp/kafka-python/pull/1750).

### Checklist

### DIFF
--- a/aiokafka/abc.py
+++ b/aiokafka/abc.py
@@ -1,5 +1,8 @@
 import abc
 from kafka import ConsumerRebalanceListener as BaseConsumerRebalanceListener
+from kafka.oauth.abstract import AbstractTokenProvider \
+    as BaseAbstractTokenProvider
+import asyncio
 
 
 class ConsumerRebalanceListener(BaseConsumerRebalanceListener):
@@ -87,6 +90,17 @@ class ConsumerRebalanceListener(BaseConsumerRebalanceListener):
         pass
 
 
+class AbstractTokenProvider(BaseAbstractTokenProvider):
+    async def token(self):
+        return await asyncio.get_running_loop().run_in_executor(
+            None, self._token)
+
+    @abc.abstractmethod
+    def _token(self):
+        pass
+
+
 __all__ = [
-    "ConsumerRebalanceListener"
+    "ConsumerRebalanceListener",
+    "AbstractTokenProvider"
 ]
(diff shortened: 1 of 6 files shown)


oulydna:
Thanks @tvoinarovskyi. Really appreciate the quick reply. 

Yes, I tested it in our own environment. 

Good point on the async callback. I will look into this.

aiokafka/conn.py
```diff
@@ -692,3 +717,47 @@ def create_salted_password(self, salt, iterations):
...
+
+        resp = yield data, True
+
+        assert resp == b"", (
+            "Server should either close or send an empty response"
+        )
+
+    def _build_oauth_client_request(self):
+        return "n,,\x01auth=Bearer {}{}\x01\x01".format(
+            self._sasl_oauth_token_provider.token(), self._token_extensions()
```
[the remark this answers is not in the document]
oulydna: shouldn't we return None based on self._token_sent?

oulydna:
@tvoinarovskyi can you please have a look at the new changes?

aiokafka/abc.py
```diff
@@ -87,6 +90,17 @@ def on_partitions_assigned(self, assigned):
         pass
 
 
+class AbstractTokenProvider(BaseAbstractTokenProvider):
+    async def token(self):
```
tvoinarovskyi, line 94: I think only providing a token async function with a proper specification is enough, let's remove the `run_in_executor`. By introducing the `_token` abstract function below you essentially require the developer to override it (you can't initialize an `abc` class without overriding all abstract methods), which is strange in the case where he only needs an async callback.
We can add a docstring describing how you can use synchronous callback instead.

aiokafka/abc.py
```diff
@@ -87,6 +90,17 @@ def on_partitions_assigned(self, assigned):
...
+        return await asyncio.get_running_loop().run_in_executor(
+            None, self._token)
+
+    @abc.abstractmethod
+    def _token(self):
+        pass
+
```
tvoinarovskyi, line 101: Please also add an `extensions` method, but leave it returning an empty sequence (not abstract method). This way the user would be able to understand how he can add extensions, but will not be required to do so.

tvoinarovskyi:
@oulydna Looks good for a merge, few minor points if you have time and let's ship it 🚢

tvoinarovskyi:
Thanks again!

oulydna:
Thanks you @tvoinarovskyi . Really glad to have this integrated. We like aiokafka, and will like it better.

PR #663, opened 2020-10-06
add test for sasl ssl

LizardWizzard:
### Changes

Hello! Thank you for great library! Just wanted to test if sasl_ssl auth method is working correctly while setting it up for my project. Realised that this scenario is missing in library's functional tests. So I've added it.

By the way, when tried to run tests on master there was a bunch of warnings like `RuntimeWarning: coroutine 'TestKafkaClientIntegration.test_failed_bootstrap' was never awaited` so many tests didnt actually run. When tried on top of the latest release tests run smoothly

### Checklist

### DIFF
--- a/tests/_testutil.py
+++ b/tests/_testutil.py
@@ -392,12 +392,21 @@ class KafkaIntegrationTestCase(unittest.TestCase):
         # Make sure there are no duplicates
         self.assertEqual(len(set(messages)), num_messages)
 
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
Thanks for the contribution! I remember having issues with the SSL+SASL scenario, because we create certificates for SSL in a very simple manner, not sure what was the specific problem.

PR #667, opened 2020-10-15
Drop Python 3.5 support, fix CI

ods:
### Changes

* Drop support for Python 3.5
* Revert switch to pytest-asyncio as it doesn't support `unittest.TestCase`
* Remove compatibility code branches for Python versions we already don't support

### Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,11 +4,7 @@
 name: Tests
 
 on:
-  push:
-    branches: [ master ]
-    tags:
-      - "v0.[0-9]+.[0-9]+"
-      - "v0.[0-9]+.[0-9]+.dev*"
+  push: {}
   pull_request:
     branches: [ master ]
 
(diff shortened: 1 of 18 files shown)


asvetlov (approved):
LGTM but please wait for a little for @tvoinarovskyi review

tvoinarovskyi:
Hi @ods, thanks for working on this.
Looks good, I am not sure how the ``python-asyncio`` addition did not affect my environment, maybe I failed to rebuild something.

ods:
> I am not sure how the `python-asyncio` addition did not affect my environment, maybe I failed to rebuild something.

`pytest.mark.asyncio` just doesn't work for `unittest.TestCase` methods. I'd like to rewrite tests pytest-style (with fixtures and functions), but that's separate task and deserves a separate PR. And after fixing warnings for Python 3.8/3.9

tvoinarovskyi:
There is a separate branch that removes loop usage called fixpy38 I think. Mostly it's finished aside from a few points that I don't remember any more :)

.github/workflows/tests.yml
```diff
@@ -4,11 +4,7 @@
 name: Tests
 
 on:
-  push:
```
tvoinarovskyi, line 7: Feel like we should revert this part. Tests are running 2ce without the limit. I think you use it to run on the fork, right? Maybe there is a better way to configure Actions...

ods: Reverted in #669

PR #672, opened 2020-10-24
Fix consumer fetcher for python3.9

dutradda:
### Changes

Fixes #
- Fix consumer fetcher for python3.9

From python3.9 the `asyncio.wait` function iterates over futures two times.
The consumer fetcher gives an iterator to this function, so the second iteration is empty, that causes the error `Set of Futures is empty`.

python3.9 asyncio code:

```python
    if any(coroutines.iscoroutine(f) for f in set(fs)):
        warnings.warn("The explicit passing of coroutine objects to "
                      "asyncio.wait() is deprecated since Python 3.8, and "
                      "scheduled for removal in Python 3.11.",
                      DeprecationWarning, stacklevel=2)

    fs = {ensure_future(f, loop=loop) for f in set(fs)}

```

### Checklist

### DIFF
--- a/CHANGES/python3.9-fix.bugfix
+++ b/CHANGES/python3.9-fix.bugfix
@@ -1,0 +1,1 @@
+Fix consumer fetcher for python3.9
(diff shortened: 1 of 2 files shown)


tvoinarovskyi:
Wow, did not think of this problem. Thanks for looking into it!

PR #676, opened 2020-10-27
Fix warnings with Python 3.9

ods:
### Changes

Python 3.9 issue warnings when coroutine is passed to `asyncio.wait()`, so we have to explicitly create task from coroutine or use future instead.


tvoinarovskyi (approved):
Lgtm!

PR #693, opened 2020-12-04
Don't pass loop explicitly in docs and examples

ods:
### Changes

* Remove docs and examples for Python 3.5
* Replace `@asyncio.coroutine` and `yield from` with `async def` and `await`
* Update docs and examples to not pass loop to `AIOKafkaConsumer`/`AIOKafkaProducer`, use `asyncio.run()`

### Checklist


benchmark/simple_consume_bench.py
```diff
@@ -116,21 +115,7 @@ def main():
...
-    loop.add_signal_handler(signal.SIGTERM, signal_hndl)
-    loop.add_signal_handler(signal.SIGINT, signal_hndl)
-
-    try:
-        loop.run_forever()
-    finally:
-        loop.close()
-        if not task.cancelled():
-            task.result()
+    asyncio.run(Benchmark(args).bench_simple())
```
tvoinarovskyi, line 118: Did you try running the benchmark? I think I had a very specific reason for writing the code that way, but it may be related to issues on earlier versions of Python.

benchmark/simple_produce_bench.py
```diff
@@ -139,21 +139,7 @@ def main():
...
-    loop.add_signal_handler(signal.SIGTERM, signal_hndl)
-    loop.add_signal_handler(signal.SIGINT, signal_hndl)
-
-    try:
-        loop.run_forever()
-    finally:
-        loop.close()
-        if not task.cancelled():
-            task.result()
+    asyncio.run(Benchmark(args).bench_simple())
```
tvoinarovskyi, line 142: Same as above

tvoinarovskyi (approved):
LGTM

benchmark/simple_consume_bench.py
ods: Just have run them and I see no difference. Probably it was the reason with old versions of asyncio.

PR #701, opened 2020-12-21
fix AIOKafkaProducer contextmanager

shamrin:
### Changes

`AIOKafkaProducer` currently doesn't work as a context manager:

```python
async with AIOKafkaProducer(loop=asyncio.get_event_loop(), **kafka_params()) as producer:
        await producer.send_and_wait(KAFKA_TOPIC, b'message data')
```

It fails with `AttributeError: 'NoneType' object has no attribute 'send_and_wait'`.

### Checklist

### DIFF
--- a/aiokafka/producer/producer.py
+++ b/aiokafka/producer/producer.py
@@ -557,6 +557,7 @@ class AIOKafkaProducer(object):
 
     async def __aenter__(self):
         await self.start()
+        return self
 
     async def __aexit__(self, exc_type, exc, tb):
         await self.stop()


shamrin:
The change it's pretty straightforward, only a single line of code. Do you think it could get merged without going through the whole checklist? `AIOKafkaProducer` is the only context manager that is broken at the moment, others works fine. For example, ``AIOKafkaConsumer`.

omikader:
+1 Can this be merged soon?

ods:
It would be nice to add test cases for all context managers

shamrin:
@ods thank you for your reply! Unfortunately, I don't have time to write the tests. I'm no longer using aiokafka actively. Somebody else is welcome to write them. @omikader? :)

In the meantime, here is the work around I've used in the past:

```python
import aiokafka

class AIOKafkaProducer(aiokafka.AIOKafkaProducer):
    """Same as aiokafka class, but can be used as a context manager

    TODO remove the class once the fix is merged: https://github.com/aio-libs/aiokafka/pull/701
    """

    async def __aenter__(self):
        await self.start()
        return self
```

omikader:
Ok take a look at #729, I don't think it needed that many changes to unit tests. The `AIOKafkaConsumer` class already had this check so I added it to the tests for the `AIOKafkaProducer`

shamrin:
Awesome, thank you @omikader!

PR #721, opened 2021-02-12
Add 3.8 and 3.9 to setup.py

ArtyomKaltovich:
Fix project description on pypi.

Should I reflect it in changelog?


ods:
Actually there is a problem with 3.8.6+ and 3.9.* due to [the bug in `wait_for`](https://bugs.python.org/issue42130). So declaring at least 3.9 support is arguable.

PR #739, opened 2021-04-23
Fix Unclosed AIOKafkaConnection when connecting to semi-broken broker

### LINKED ISSUE #715: Connection should be close if there is a failure during start
**Describe the bug**

After calling `await consumer.start()`, `AioKafkaConnection.__del__` calls `self._loop.call_exception_handler` after failing to connect to a Kafka node. If there is a failure during startup, but after the client is "connected", aiokafka should close or cleanup the connection properly without warning about an unclosed connection and calling the loop exception handler. For example, if there was a failure during version_lookup I believe AIOKafkaConnection would be "connected", but the user code does not have a chance to close the connection. 

```
Unable connect to node with id 2:
Unclosed AIOKafkaConnection
conn: <AIOKafkaConnection host=kubeprod.aws.aurotfp.com port=31092>
Exception ignored in: <function AIOKafkaConnection.__del__ at 0x7fa2da444700>
  File "/usr/local/lib/python3.8/dist-packages/aiokafka/conn.py", line 202, in __del__
    self._loop.call_exception_handler(context)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 1744, in call_exception_handler
    self._exception_handler(self, context)
  File "/usr/local/lib/python3.8/dist-packages/featuremine/system/select_events.py", line 162, in exception_handler
    raise SystemExit("Exception occured.")
SystemExit: Exception occured.
```

aiokafka               0.7.0
kafka-python           2.0.2

bitphage:
### Changes

Fixes #715

### Checklist


aiokafka/conn.py
```diff
@@ -94,7 +94,12 @@ async def create_conn(
         sasl_kerberos_domain_name=sasl_kerberos_domain_name,
         sasl_oauth_token_provider=sasl_oauth_token_provider,
         version_hint=version_hint)
-    await conn.connect()
+    try:
+        await conn.connect()
+    except Exception:
+        # Cleanup to prevent `Unclosed AIOKafkaConnection` if we failed to connect here
+        conn.close()
```
ods, line 101: Shouldn't be this logic inside `AIOKafkaConnection.connect()`, so that it maintains invariant too?

tests/test_conn.py
```diff
@@ -247,6 +247,34 @@ async def test_close_disconnects_connection(self):
...
+        # We cannot catch ResourceWarning from __del__, so just monitor `close()` call
+        close_called = False
+
+        def mock_close(slf):
+            slf._reader = None
+            nonlocal close_called
+            close_called = True
+
+        with mock.patch.object(AIOKafkaConnection, 'connect', mock_connect):
+            with mock.patch.object(AIOKafkaConnection, 'close', mock_close):
```
ods, line 274: Is it possible to reproduce the problem without mocking?

aiokafka/conn.py
bitphage: Not sure about that, because `AIOKafkaConnection.connect()` is not being used directly, only via `create_conn` factory. Though, I can move this wrapping logic if you think that it should be more appropriate to have it there.

tests/test_conn.py
bitphage: I don't know how exactly kafka broker can be brought up to exactly this broken state unfortunately. We've got this situation multiple times on Amazon MSK when kafka cluster was under high load, which caused failures on some brokers + "under replicated partitions" issues.

ods: Without it we can be never sure, that we actually fixed the problem. Any script that eventually fails is fine, not necessary suitable for unit tests one.

bitphage: If you remove exception handling from `create_conn` and run `python -m pytest -v --log-cli-level=INFO -s --docker-image aiolibs/kafka:2.12_2.2.2 tests/test_conn.py::ConnIntegrationTest::test_semi_broken_connection` there will be original error logged in live log as:

```
Exception ignored in: <function AIOKafkaConnection.__del__ at 0x7f01a90fc3b0>
  File "/home/vvk/devel/aiokafka/aiokafka/conn.py", line 194, in __del__
    source=self)
ResourceWarning: Unclosed AIOKafkaConnection <AIOKafkaConnection host=127.0.0.1 port=43803>
```

So the unittest is still mimics the issue. Basically any exception from `connect()` after `self._read_task` is set is a problem.

ods: Are there other possible race conditions here? Like initializing `self._reader`, but not `self._read_task`?

bitphage: No I think not. The most possible source of exceptions is later call `await self._do_version_lookup()`

ods:
Looks like it was already fixed in #810

PR #757, opened 2021-06-24
Add wheel build support for linux aarch64

odidev:
Added wheel build support for linux aarch64.
Related to #756. @ods Could you please review this PR?

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,13 +36,21 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+        arch: [auto]
+        include:
+          - os: ubuntu-latest
+            arch: aarch64
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
           python-version: 3.7
+      - name: Set up QEMU
+        if: ${{ matrix.arch == 'aarch64' }}
+        uses: docker/setup-qemu-action@v1
       - name: Build wheels
         env:
+          CIBW_ARCHS_LINUX: ${{matrix.arch}}
           CIBW_BUILD: cp36-* cp37-* cp38-* cp39-*
           CIBW_BEFORE_BUILD_LINUX: pip install -r requirements-cython.txt && yum install -y zlib-devel
           # On windows and mac we should have z library preinstalled


ods:
Is it for Linux only? I'd expect it for Mac OS too. Don't we need add test pipelines on this arch too? Have you tested the wheel produced with this pipeline?

odidev:
I ran tests for the aarch64 wheels and getting the following error: 
https://github.com/odidev/aiokafka/runs/5121100722?check_suite_focus=true#step:5:564 

@ods, Could you please have a look and share you suggestion on the same?

ods:
> I ran tests for the aarch64 wheels and getting the following error: https://github.com/odidev/aiokafka/runs/5121100722?check_suite_focus=true#step:5:564

The error is [from SSL_CTX_use_certificate_chain_file call](https://github.com/python/cpython/blob/v3.8.12/Modules/_ssl.c#L4033-L4046).

> A return code equal to 0 indicates an error. The following are the most likely causes of errors:
> - The certificate file does not exist or you do not have permission to read that file.
> - The certificate file that contains the certificate chain is not in PEM format.
> - If you loaded a private key file before issuing this function, the private key in that file does not match the corresponding public key in the certificate.

The file is here and seems to be valid.

ods:
It would be nice to log openssl version (`ssl.OPENSSL_VERSION_INFO`).

odidev:
Output for `ssl.OPENSSL_VERSION_INFO`:
https://github.com/odidev/aiokafka/runs/5156311151?check_suite_focus=true#step:5:553

ods:
```
OPENSSL_VERSION_INFO= (1, 1, 1, 13, 15)
```
So, it's not due to update to openssl 3. Sorry, I have no other ideas besides checking all possible reasons, like: log current process' uid/gid and uid/gid and permissions of the file, try to read file and log its content, try to check the file with `openssl x509 -in tests/ssl_cert/cl_client.pem -noout -text` command and log the result, etc. The best way would be to reproduce it locally, but I have no access to device with m1 chip.

odidev:
The command `openssl x509 -in tests/ssl_cert/cl_client.pem -noout –text` is giving the following output on both x86_64 and arm64 machines: 

```
root@6fc8203969dc:/aiokafka# openssl x509 -in tests/ssl_cert/cl_client.pem -noout -text 
unable to load certificate 
281473578916320:error:0909006C:PEM routines:get_name:no start line:../crypto/pem/pem_lib.c:745:Expecting: TRUSTED CERTIFICATE 
```

ods:
Merged in https://github.com/aio-libs/aiokafka/pull/839

PR #782, opened 2021-09-09
Fix KeyError on solitary abort marker.

### LINKED ISSUE #781: Corner case: Aborted transaction yields KeyError in fetcher.py
**Describe the bug**
At least with recent versions if the `bitnami/kafka:2` Kafka broker container image, aborted transactions can give rise to a situation in which the log for a given Kafka topic looks as follows:

```sh
$ /opt/bitnami/kafka/bin/kafka-dump-log.sh --files /bitnami/kafka/data/test-topic-0/00000000000000000000.log
Dumping /bitnami/kafka/data/test-topic-0/00000000000000000000.log
Starting offset: 0
baseOffset: 0 lastOffset: 0 count: 1 baseSequence: -1 lastSequence: -1 producerId: 3 producerEpoch: 1 partitionLeaderEpoch: 0 isTransactional: true isControl: true position: 0 CreateTime: 1631000770035 size: 78 magic: 2 compresscodec: NONE crc: 2811855458 isvalid: true
```

Here, the log consists of the transaction abort marker _only_. Therefore, https://github.com/aio-libs/aiokafka/blob/dbd490fb9b015a018f238bfb5574687e3c7f74ce/aiokafka/consumer/fetcher.py#L267 is never reached and https://github.com/aio-libs/aiokafka/blob/dbd490fb9b015a018f238bfb5574687e3c7f74ce/aiokafka/consumer/fetcher.py#L223 results in a `KeyError` when attempting to consume the aforementioned topic.

**Expected behaviour**
Consumer should proceed without error.

 - aiokafka version (`python -c "import aiokafka; print(aiokafka.__version__)"`): 0.7.2
 - kafka-python version (`python -c "import kafka; print(kafka.__version__)"`): 2.0.2
 - Kafka Broker version (`kafka-topics.sh --version`): 2.8.0 (Commit:ebb1d6e21cc92130)
 - Other information (Confluent Cloud version, etc.):

**Reproducible example**
```python
import logging
import asyncio

async def _recv_all():
    consumer = AIOKafkaConsumer(
        bootstrap_servers=['172.30.0.3'],
        client_id=f'test123-test-consumer',
        auto_offset_reset="earliest",
        isolation_level="read_committed",
        enable_auto_commit=False
    )
    async with consumer:
        consumer.subscribe(['test-topic'])
        while True:
            for messages in (await consumer.getmany(
                    timeout_ms=20)).values():
                if messages:
                  print('Got message')


logging.basicConfig(level=logging.DEBUG)
loop = asyncio.new_event_loop()
result = loop.run_until_complete(_recv_all())
loop.run_until_complete(loop.shutdown_asyncgens())
return result
```

pikulmar:
### Changes

Fixes #781 

### Checklist

### DIFF
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -5,7 +5,9 @@ from unittest import mock
 
 from kafka.protocol.offset import OffsetResponse
 from aiokafka.record.legacy_records import LegacyRecordBatchBuilder
-from aiokafka.record.default_records import DefaultRecordBatchBuilder
+from aiokafka.record.default_records import (
+    # NB: test_solitary_abort_marker relies on implementation details
+    _DefaultRecordBatchBuilderPy as DefaultRecordBatchBuilder)
 from aiokafka.record.memory_records import MemoryRecords
 
 from aiokafka.protocol.fetch import (
--- a/tests/test_fetcher.py
+++ b/tests/test_fetcher.py
@@ -5,6 +5,8 @@ from unittest import mock
 
 from kafka.protocol.offset import OffsetResponse
 from aiokafka.record.legacy_records import LegacyRecordBatchBuilder
+from aiokafka.record.default_records import DefaultRecordBatchBuilder
+from aiokafka.record.memory_records import MemoryRecords
 
 from aiokafka.protocol.fetch import (
     FetchRequest_v0 as FetchRequest, FetchResponse_v0 as FetchResponse)
@@ -18,7 +20,7 @@ from aiokafka.structs import (
 from aiokafka.client import AIOKafkaClient
 from aiokafka.consumer.fetcher import (
     Fetcher, FetchResult, FetchError, ConsumerRecord, OffsetResetStrategy,
-    PartitionRecords, READ_UNCOMMITTED
+    PartitionRecords, READ_COMMITTED, READ_UNCOMMITTED
 )
 from aiokafka.consumer.subscription_state import SubscriptionState
 from aiokafka.util import create_future, create_task, get_running_loop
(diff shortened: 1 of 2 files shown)


tests/test_fetcher.py
```diff
@@ -5,6 +5,9 @@
 
 from kafka.protocol.offset import OffsetResponse
 from aiokafka.record.legacy_records import LegacyRecordBatchBuilder
+from aiokafka.record.default_records import DefaultRecordBatchBuilder
+from aiokafka.record.memory_records import MemoryRecords
+from aiokafka.consumer.fetcher import PartitionRecords
```
ods, line 10: Duplicate import

pikulmar: Fixed in 49c4be4

pikulmar: Apologies for force-pushing, did not notice that you already started the review.

tests/test_fetcher.py
```diff
@@ -534,3 +536,27 @@ async def mock_send(node_id, request):
...
+            tp=TopicPartition('test-topic', 0),
+            records=records,
+            aborted_transactions=[],
+            fetch_offset=0,
+            key_deserializer=None,
+            value_deserializer=None,
+            check_crcs=True,
+            isolation_level=READ_COMMITTED)
+
+        self.assertEqual(len(list(partition_recs)), 0)
```
ods, line 562: This test doesn't fail if I switch back to `self._aborted_producers.remove()` when I run it locally. Does it work for you?

pikulmar: You are right, let me try to figure out why that is...

pikulmar: I updated the test in d6cfb88. The problem was that the test message was not correctly constructed.

pikulmar: Hm, it seems that the workaround for creating the control batch does not work with C extension. Do you have any suggestions how to best address that?

pikulmar: I took a somewhat lazy approach to address this in 35a843a. Let me know what you think is the "right" way to go about this.

ods: I still don't get what you're trying to do in test. Could you explain it please, better directly in comments?  
There is always a naïve approach to disable the test when it's run with C extension (use `aiokafka.util.NO_EXTENSIONS`), but I'd try to find a way that works for both cases.

ods: > I took a somewhat lazy approach to address this in 35a843a

It looks ok to me. But explanation in comments would be appreciated anyway.

pikulmar: Added explanatory comments in e418952.

ods (approved):

PR #789, opened 2021-10-22
Fix authentication log message when using SCRAM sasl_mechanism

ran-ka:
### Changes

Really small logging change that confused me for a long time, until I realized it's a bug :-)


ods:
Already fixed in #861

PR #810, opened 2021-12-29
Fix warnings

ods:
### Changes

Fixe various warnings (ResourceWarning, DeprecationWarning).

### Checklist

### DIFF
--- a/aiokafka/conn.py
+++ b/aiokafka/conn.py
@@ -227,11 +227,15 @@ class AIOKafkaConnection:
             self._idle_handle = loop.call_soon(
                 self._idle_check, weakref.ref(self))
 
-        if self._version_hint and self._version_hint >= (0, 10):
-            await self._do_version_lookup()
+        try:
+            if self._version_hint and self._version_hint >= (0, 10):
+                await self._do_version_lookup()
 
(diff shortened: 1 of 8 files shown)


aiokafka/conn.py
```diff
@@ -227,11 +227,15 @@ async def connect(self):
...
-            await self._do_version_lookup()
+        try:
+            if self._version_hint and self._version_hint >= (0, 10):
+                await self._do_version_lookup()
 
-        if self._security_protocol in ["SASL_SSL", "SASL_PLAINTEXT"]:
-            await self._do_sasl_handshake()
+            if self._security_protocol in ["SASL_SSL", "SASL_PLAINTEXT"]:
+                await self._do_sasl_handshake()
+        except:  # noqa: E722
```
tvoinarovskyi, line 236: Please change this to ``except Exception``, I would prefer not to have an implicit catch on BaseException if there is no need for it.

tvoinarovskyi (approved):

aiokafka/conn.py
ods: We do need to catch BaseException. Actually I stepped into the problem with CancelledError, which nowadays doesn't inherit from Exception.

PR #813, opened 2021-12-31
Fix Appveyor builds, simplify Windows builds

multani:
### Changes

Appveyor builds should now be passing: https://ci.appveyor.com/project/multani/aiokafka/history

* Appveyor builds don't work since several months due to the unsupported `--no-print-logs` pytest option.

* Appveyor builds were using outdated/old Kakfa & Python versions

* Windows builds were pre-installing the python-snappy wheel ... [for another version to be installed from `requirements-win-test.txt`](https://github.com/aio-libs/aiokafka/blob/654fcbf6ac9d4ec8edc6c26bd1500383113fc14d/requirements-win-test.txt#L12) just after.

About Appveyor: since [GitHub Actions is already building on Windows](https://github.com/aio-libs/aiokafka/blob/654fcbf6ac9d4ec8edc6c26bd1500383113fc14d/.github/workflows/tests.yml#L70-L146), Appveyor could be simply completely removed.

In this case, it would be cool to also quickly have a look in the project settings on GitHub to remove references to Appveyor (hooks, etc.)

### Checklist

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,21 +1,20 @@
 environment:
   PYTHONASYNCIODEBUG: "1"
-  KAFKA_VERSION: "0.10.2.1"
-  SCALA_VERSION: "2.11"
+  KAFKA_VERSION: "2.8.1"
+  SCALA_VERSION: "2.13"
   matrix:
-    - PYTHON: "C:\\Python36"
-      SNAPPY_WHEEL: "tools\\python_snappy-0.5.4-cp36-cp36m-win32.whl"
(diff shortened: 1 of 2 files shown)


ods (approved):

tvoinarovskyi:
@multani Thanks, seems like snappy is now installing the newer version correctly on Windows. I see they added wheels for Windows, which makes things much easier. Before I had to download separately build wheel packages because Appveyor/Actions could not build snappy from the source.

I think removing Appveyor is the correct approach moving forward, as we moved to Github actions now.

Great thanks for the contributions!

PR #832, opened 2022-05-06
fix flake check

aoberegg:
### Changes

Adjusts dependencies in order to be compatible with newest click version (https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click)

image

### Checklist
This updates only dependencies.

### DIFF
--- a/requirements-ci.txt
+++ b/requirements-ci.txt
@@ -1,6 +1,6 @@
 -r requirements-cython.txt
 flake8==4.0.1
-black==21.10b0
+black>=21.10b0
 mypy==0.910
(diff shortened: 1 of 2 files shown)


requirements-ci.txt
```diff
@@ -1,6 +1,6 @@
 -r requirements-cython.txt
 flake8==4.0.1
-black==21.10b0
+black>=21.10b0
```
ods, line 3: According to your reference is should be `>=22.3.0`

aoberegg: Thanks for the hint. I fixed it.

Issue #842, opened 2022-07-01
Why ConsumerRebalanceListener is not expose when create a Consumer?


marcosschroh:
Hi,

Thanks for the work that you are doing!! 

According to the code, the only way to use the `ConsumerRebalanceListener` is using the `subscribe` method, I was wondering: Why is not also possible to set a `ConsumerRebalanceListener` when a `Consumer` is created using the `__init__`? I think if we do this, the testing will become easier because of mocking and you won’t be force to use the `subscribe` method. 

In the `init` method is used 
`self._subscription.subscribe(…)`  which accepts a `ConsumerRebalanceListener` argument, so I think it is possible to make it. 

Ideally, I think both options should be possible. If the user set a new `ConsumerRebalanceListenter` using the `subscribe` method, it should prevail.

what do you think?

PR #845, opened 2022-07-06
feat: rebalance listener added to Consumer constructor

marcosschroh:
### Changes

Fixes https://github.com/aio-libs/aiokafka/issues/842

* Now it is possible to set the `ConsumerRebalanceListener`  when a `Consumer` instance is created.
* The validation is made in the `SubscriptionState._validate_rebalance_listener` method.
* The `RebalanceListenerCT` (TypeVar) was created. Using this, we can say that the listener should be a subtype of the abstract `ConsumerRebalanceListener` 
* `sphinx_autodoc_typehints` extension was added in order to make possible the `TypeVar` documentation. If you prefer not to use `TypeVar` we can remove this and the `RebalanceListenerCT`
* Some typing added

### Checklist

- [x] Unit tests for the changes exist (The existing test covers the new changes)


aiokafka/abc.py
```diff
@@ -1,6 +1,11 @@
 import abc
+import typing
+
 from kafka import ConsumerRebalanceListener as BaseConsumerRebalanceListener
 
+RebalanceListenerCT = typing.TypeVar(
+    "RebalanceListenerCT", bound="ConsumerRebalanceListener")
```
ods, line 7: Why do you create a TypeVar here? We don't have generics where it's used.

aiokafka/consumer/consumer.py
```diff
@@ -1046,18 +1074,14 @@ def subscribe(self, topics=(), pattern=None, listener=None):
...
         if not (topics or pattern):
             raise ValueError(
                 "You should provide either `topics` or `pattern`")
         if topics and pattern:
             raise ValueError(
                 "You can't provide both `topics` and `pattern`")
-        if listener is not None and \
-                not isinstance(listener, ConsumerRebalanceListener):
-            raise TypeError(
-                "listener should be an instance of ConsumerRebalanceListener")
```
ods, line 1060: This check is still correct, why removing it?

tests/test_consumer.py
```diff
@@ -132,6 +132,15 @@ def test_create_consumer_no_running_loop(self):
             loop.run_until_complete(consumer.stop())
             loop.close()
 
+    @run_until_complete
+    async def test_create_consumer_with_listener(self):
+        listener = StubRebalanceListener()
```
ods, line 137: There are 2 tests here, it would be better to separate them. 
And the both test check validation only. But we also have to check that custom listener is actually used, can we do that too?

tvoinarovskyi:
Hi @marcosschroh,
Thank you for the contribution, could you elaborate as to why would it be essential to have a listener on creation time? As you can guess most listeners would require the back link to Consumer class, so as per Java's API we modeled the subscribe method to be the only place this can be passed. I feel like changing the API to set "consumer" is not completely backward compatible and I would like to have a good reason to include it. Thanks!

tvoinarovskyi:
For example, there can be code, where listener subclasses have "consumer" property, which is not settable.

marcosschroh:
Hi @tvoinarovskyi 

Thanks for taking a look. I see 2 problems here:

1. Why do I need to always use the `Consumer.subscribe` method if I want to use a `RebalanceListener`? It is a bit weird because I can consume immediately after creating a `Consumer`. In fact, the `Consumer.__init__` method calls the `_subscription.subscribe(...)` which accepts a `RebalanceListener`, so it should be possible to set a `listener` during consumer creation time.  

As you said:
> most listeners would require the back link to Consumer class

I could see that when I saw the examples and tests. Most of the cases (probably always) you need a link to the consumer from the `RebalanceListener` instance, which made me realized that is only possible to set a `RebalanceListener` using the `subscribe` because the consumer instance is required (chicken/egg). This is leading to the second problem/question:

2. Looks like the class `aiokafka.abc.ConsumerRebalanceListener` should have a `def  __init__(Optional[Consumer])` method. In all the test cases and examples the same code is repeated:

```python
 class RebalanceListener(ConsumerRebalanceListener):

        def __init__(self, consumer):  # Always repeat the same
            self.consumer = consumer
        ...
```

There we can see that the end user always has to include the `__init__(consumer)` method. 

Another problem (out of this scope) is that the interface defines `sync` methods, but in examples and test cases always `async` methods are defined (`async def on_partitions_revoked` and `async def on_partitions_assigned`).

This `PR` adds the `def __init__(Optional[consumer] = None): ...` to the class `aiokafka.abc.ConsumerRebalanceListener`, so end users won't have to define the `RebalanceListener.__init__(consumer): ...`  every time . Second, it will be possible to set a `listener` on consumer creation time as you have a reference to `self` (Consumer instance).

This change shouldn't affect current users because they are already defining the `RebalanceListener.__init__(consumer)` and it will made things easier for new users that want to use a `RebalanceListener`. The `consumer` property will be always present, regardless whether the listener was set during consumer creation time or using the `subscribe` method.

I hope my explanation makes sense.

PR #858, opened 2022-09-08
docs: remove incorrect await

yi-jiayu:
### Changes

`AIOKafkaConsumer.highwater` is not an async method:

https://github.com/aio-libs/aiokafka/blob/6360747c4305d1d096d5e55037d60a12c4c9ab15/aiokafka/consumer/consumer.py#L668

Remove an `await` before a call to `consumer.highwater` in an example in the docs.


wbarnha:
@webknjaz documentation correction, minor fix

Issue #852, opened 2022-07-27
SCRAM SASL mechanism wrongly logged as PLAIN


b4dboi:
Hi,

I found out that the SCRAM SASL mechanism is likely forgotten in the logging logic at conn.py. I am using a SCARAM-SHA-256 SASL mechanism and after a client successfully connects to Kafka the log shows `Authenticated as xxx via PLAIN`. 

I believe there is a missing condition `elif self._sasl_mechanism.startswith('SCRAM-SHA-'):` before the following line. https://github.com/aio-libs/aiokafka/blob/6360747c4305d1d096d5e55037d60a12c4c9ab15/aiokafka/conn.py#L340

PR #861, opened 2022-10-14
handled other SASL mechanism in logging

mangin:
Handles other sasl_mechanism into logging like:
'PLAIN', 'SCRAM-SHA-256', 'SCRAM-SHA-512'

Now we always write that we use PLAIN as auth sasl_mechanism.


mangin:
The change is so small. So i didn't update changelog

ods:
> The change is so small. So i didn't update changelog
Please create `CHANGES/852.bugfix` file anyway.

Fixes #852.

mangin:
>Please create CHANGES/852.bugfix file anyway.

Okay

PR #870, opened 2022-12-13
Add annotation for consumer

shizacat:
For more smoothly work in IDE, please will add annotation to 'getone' and 'getmany' methods.


wbarnha:
@webknjaz safe merge, just adds type annotations.

PR #890, opened 2023-05-26
Support SASL authentication in AIOKafkaAdminClient

### LINKED ISSUE #889: AIOKafkaAdminClient does not accept SASL authentication kwargs
**Describe the bug**

`AIOKafkaAdminClient` does not accept SASL extra authentication keyword arguments, required for authentication in Amazon MSK.

When trying to pass them, it fails with the following error:

```
  File "kafka_bug_mre.py", line 4, in <module>
    admin_client = AIOKafkaAdminClient(
TypeError: AIOKafkaAdminClient.__init__() got an unexpected keyword argument 'sasl_mechanism'
```

**Expected behaviour**

`AIOKafkaAdminClient.__init__` should accept all arguments  that are valid for a simple `AIOKafkaClient`.

**Environment:**
 - aiokafka version 0.8.0
 - kafka-python version: 2.0.2
 - Kafka Broker version 7.3.2-ccs (Commit:853191ff421b2935dfa531545651ab667b809801)

**Reproducible example**
```python
from aiokafka.admin import AIOKafkaAdminClient
from aiokafka.helpers import create_ssl_context

admin_client = AIOKafkaAdminClient(
    bootstrap_servers=["localhost:9092"],
    sasl_mechanism="SCRAM-SHA-512",
    sasl_plain_username="username",
    sasl_plain_password="password",
    security_protocol="SASL_SSL",
    ssl_context=create_ssl_context(),
)
```

selevit:
### Changes

Fixes #889

Added the following keyword arguments into `AioKafkaAdminClient`:

1. `sasl_mechanism`.
2. `sasl_plain_username`
3. `sasl_plain_password`
4. `sasl_kerberos_service_name`
5. `sasl_kerberos_domain_name`
6. `sasl_oauth_token_provider`


ods:
It would be nice to cover them with tests.

selevit:
> It would be nice to cover them with tests.

@ods, TBH it's not quite clear, why the coverage is decreased so much. Especially taking into account the there shouldn't be any coverage loss because of this change - parameters are just passed to `AioKafkaClient` as is.

I agree that would be nice to cover the SASL auth by tests in general. But it implies configuring a separate kafka instance with those settings, and IMO this should have been done when adding this SASL functionality into the main client.

Regarding this particular change, I just proxy the same parameters into `AioKafkaClient` and that's it. This code should run it tests, because there are tests that create `AioKafkaAdminClient` class, and parameters should be passed with their default values. Yes, the SASL functionality itself is not covered, is it actually covered at all for the generic client?

ods:
Sure, we can introduce such simple changes with just manual testing. But producer/consumer have the tests for some of SASL mechanisms in `tests/test_sasl.py`. Doesn't the same approach work for admin client?

selevit:
Oh, thanks for pointing them out.

I have added a first testcase into test_sasl.py: `test_admin_client_sasl_plaintext_basic`. But when running this, I get an error like this:

```cpp
kafka.errors.IncompatibleBrokerVersion: IncompatibleBrokerVersion: Kafka broker does not support the 'MetadataRequest_v0' Kafka protocol.
```

How to reproduce:
```
pytest --log-cli-level=INFO -s --docker-image aiolibs/kafka:2.13_2.8.1 tests/test_sasl.py -k test_admin_client_sasl_plaintext_basic
```

I noticed that there was a [similar bug](https://github.com/dpkp/kafka-python/issues/1924) in python-kafka, but it's fixed in 2.0.2 already.

Any quick idea about the reason? Exactly the same calls work in `test_admin.py` but with non-sasl brokers.

PR #912, opened 2023-08-01
912.bugfix allow checksum to be `None` on `ConsumerRecord`

zschumacher:
### Changes

When using serialization frameworks like pydantic, I can not attach a consumer record because checksum can be None.  Since it is marked as deprecated, it seems reasonable to mark it as optional here.

### Checklist


zschumacher:
> Add a new news fragment into the CHANGES folder

Not really sure what this means - do I need to create a CHANGES folder?

PR #827, opened 2022-03-17
Implement KIP-345 in aiokafka (rebase of #682) 

ntextreme3:
### Changes

Fixes #680 

This is a rebase of https://github.com/aio-libs/aiokafka/pull/682 -- see https://github.com/aio-libs/aiokafka/pull/682#issuecomment-1070337629

### Checklist

- [ ] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

PR #941, opened 2023-11-15
Implement KIP-345 in aiokafka (rebase of #827)

### LINKED ISSUE #680: Support for static membership protocol
Implement support for static group membership to avoid unnecessary rebalances
[KIP-345](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances)

**Additional context**
It looks like this feature is implemented in a fork of this library
https://github.com/robinhood/aiokafka/pull/21

joshuaherrera:
### Changes

Fixes #680

This is a rebase of #827 that adds test for KIP-345 functionality - See [#827 (comment)](https://github.com/aio-libs/aiokafka/pull/827#issuecomment-1783797327)

### Checklist

### DIFF
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -8,8 +8,8 @@ class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
     """
 
     @abc.abstractmethod
-    def assign(self, cluster, members,
-               member_group_instance_ids):  # lgtm[py/inheritance/signature-mismatch]
+    def assign(self, cluster, members,  # lgtm[py/inheritance/signature-mismatch]
+               member_group_instance_ids):
         """Perform group assignment given cluster metadata, member subscriptions
            and group_instance_ids
         Arguments:
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -15,8 +15,8 @@ class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
             cluster (ClusterMetadata): metadata for use in assignment
             members (dict of {member_id: MemberMetadata}): decoded metadata for
                 each member in the group.
-            member_group_instance_ids members (dict of {member_id: MemberMetadata}): decoded
-                metadata for each member in the group.
+            member_group_instance_ids (dict of {member_id: MemberMetadata}):
+                decoded metadata for each member in the group.
         Returns:
             dict: {member_id: MemberAssignment}
         """
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -8,7 +8,8 @@ class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
     """
 
     @abc.abstractmethod
-    def assign(self, cluster, members, member_group_instance_ids): # lgtm[py/inheritance/signature-mismatch]
+    def assign(self, cluster, members,
+               member_group_instance_ids):  # lgtm[py/inheritance/signature-mismatch]
         """Perform group assignment given cluster metadata, member subscriptions
            and group_instance_ids
         Arguments:
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -15,8 +15,8 @@ class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
             cluster (ClusterMetadata): metadata for use in assignment
             members (dict of {member_id: MemberMetadata}): decoded metadata for
                 each member in the group.
-            mmember_group_instance_ids members (dict of {member_id: MemberMetadata}): decoded metadata for
-                each member in the group.
+            member_group_instance_ids members (dict of {member_id: MemberMetadata}): decoded
+                metadata for each member in the group.
         Returns:
             dict: {member_id: MemberAssignment}
         """
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -1,0 +1,23 @@
+from kafka.coordinator.assignors.abstract import AbstractPartitionAssignor
+import abc
+
+class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
+    """
+    Abstract assignor implementation that also supports static assignments (KIP-345)
+    """
+
+
+    @abc.abstractmethod
+    def assign(self, cluster, members, member_group_instance_ids):
+        """Perform group assignment given cluster metadata, member subscriptions
+           and group_instance_ids
+        Arguments:
+            cluster (ClusterMetadata): metadata for use in assignment
+            members (dict of {member_id: MemberMetadata}): decoded metadata for
+                each member in the group.
+            mmember_group_instance_ids members (dict of {member_id: MemberMetadata}): decoded metadata for
+                each member in the group.
+        Returns:
+            dict: {member_id: MemberAssignment}
+        """
+        pass
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -8,7 +8,7 @@ class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
     """
 
     @abc.abstractmethod
-    def assign(self, cluster, members, member_group_instance_ids):
+    def assign(self, cluster, members, member_group_instance_ids): # lgtm[py/inheritance/signature-mismatch]
         """Perform group assignment given cluster metadata, member subscriptions
            and group_instance_ids
         Arguments:
--- a/aiokafka/consumer/assignors.py
+++ b/aiokafka/consumer/assignors.py
@@ -1,11 +1,11 @@
 from kafka.coordinator.assignors.abstract import AbstractPartitionAssignor
 import abc
 
+
 class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
     """
     Abstract assignor implementation that also supports static assignments (KIP-345)
     """
-
 
     @abc.abstractmethod
     def assign(self, cluster, members, member_group_instance_ids):
--- a/aiokafka/consumer/group_coordinator.py
+++ b/aiokafka/consumer/group_coordinator.py
@@ -9,13 +9,15 @@ from kafka.coordinator.protocol import ConsumerProtocol
 from kafka.protocol.commit import (
     OffsetCommitRequest_v2 as OffsetCommitRequest,
     OffsetFetchRequest_v1 as OffsetFetchRequest)
-from kafka.protocol.group import (
+from aiokafka.protocol.group import (
     HeartbeatRequest, JoinGroupRequest, LeaveGroupRequest, SyncGroupRequest)
 
 import aiokafka.errors as Errors
 from aiokafka.structs import OffsetAndMetadata, TopicPartition
 from aiokafka.client import ConnectionGroup, CoordinationType
 from aiokafka.util import create_future, create_task
+from aiokafka.consumer.assignors import AbstractStaticPartitionAssignor
+
 
 log = logging.getLogger(__name__)
 
@@ -210,6 +212,7 @@ class GroupCoordinator(BaseCoordinator):
 
     def __init__(self, client, subscription, *,
                  group_id='aiokafka-default-group',
+                 group_instance_id=None,
                  session_timeout_ms=10000, heartbeat_interval_ms=3000,
                  retry_backoff_ms=100,
                  enable_auto_commit=True, auto_commit_interval_ms=5000,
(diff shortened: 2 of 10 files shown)


joshuaherrera:
@ods Here's the PR for the tests, I'll work on getting merge conflicts fixed on my end since it's only two files, hopefully by EOD today.

aiokafka/consumer/group_coordinator.py
```diff
@@ -401,8 +407,18 @@ async def _perform_assignment(
             'Invalid assignment protocol: %s' % assignment_strategy
         member_metadata = {}
         all_subscribed_topics = set()
-        for member_id, metadata_bytes in members:
+        group_instance_id_mapping = {}
+        # for member_id, group_instance_id, metadata_bytes in members:
+        for member in members:
+            if len(member) == 2:
```
ods, line 413: This won't work well when we add typing. Could we switch by `isinstance()` instead?

aiokafka/consumer/group_coordinator.py
```diff
@@ -401,8 +407,18 @@ async def _perform_assignment(
...
+                member_id, metadata_bytes = member
+                group_instance_id = None
+            elif len(member) == 3:
+                member_id, group_instance_id, metadata_bytes = member
+            else:
+                raise Exception("unknown protocol returned from assignment")
             metadata = ConsumerProtocol.METADATA.decode(metadata_bytes)
+            group_instance_id_mapping = group_instance_id
```
ods, line 421: Probably a type here

aiokafka/consumer/assignors.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class AbstractStaticPartitionAssignor(AbstractPartitionAssignor):
+    """
+    Abstract assignor implementation that also supports static assignments (KIP-345)
+    """
+
+    @abc.abstractmethod
+    def assign(self, cluster, members,  # lgtm[py/inheritance/signature-mismatch]
+               member_group_instance_ids):
```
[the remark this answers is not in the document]
joshuaherrera: Based on the [KIP-345 proposal](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances), I don't think this class is actually necessary, since the client side need only provide the `group.instance.id` with the appropriate `Join/Sync request/responses` to utilize static membership.

With that said, I'm removing this class. I tested locally and the tests still pass.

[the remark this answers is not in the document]
joshuaherrera: Small update, I looked at the original commits over at the Robinhood fork, and it looks like this class was added as an interface for containerized deployments, so that others could extend the class and use the `group.instance.id` mapping in their implementation. See the [original commit message here, specifically paragraph three](https://github.com/robinhood/aiokafka/commit/cd44bdcf57b4aa6f675783ed1043b2d643b0ac54).

With that said, it may be worth keeping so anyone that is using an older fork to utilize KIP-345 could move over to this version of `aiokafka` with ease (in the case that they implement their own StaticPartitionAssignor.)

@ods What do you think? Any objections to keeping this?

aiokafka/consumer/group_coordinator.py
joshuaherrera: I don't think so, since `members` is a member of the `JoinGroupResponse` class [as shown here](https://github.com/joshuaherrera/aiokafka/blob/kip-345/aiokafka/protocol/group.py#L50-L55) and it is defined as an Array. Any version under v5 will be an array of 2, v5 and up currently are an array of 3.

I can add a comment for documentation purposes, so that future users know why we check for a length of two or three, unless you have another suggestion?

joshuaherrera: Another thing I just thought of was modifying the `_perform_assignment` [function signature ](https://github.com/joshuaherrera/aiokafka/blob/kip-345/aiokafka/coordinator/base.py#L189)to just accept the `JoinGroupResponse` as it's only parameter, so we can check the class in this function. So instead the function signature would be 
```python
def _perform_assignment(self, response: JoinGroupResponse):
```    
and we could then use `isinstance` to check the version of the `response` that is being used.

 Another option could be adding the [`API_VERSION`](https://github.com/joshuaherrera/aiokafka/blob/kip-345/aiokafka/protocol/group.py#L41) as a new parameter to the function, so we could just check against that here. So the function signature would become 
 ```python
 def _perform_assignment(self, leader_id, protocol, members, api_version):
 ```

aiokafka/consumer/assignors.py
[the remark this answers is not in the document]
ods: I'd prefer to keep it if it's useful. But we'll have to solve the problem with interface mismatch. I think we could move it to separate PR to not block the rest.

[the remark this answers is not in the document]
joshuaherrera: Agreed, this can be re-added in another PR to not block the rest of this work. I'll leave the PR as-is for now, since I already removed the code block from this PR.

aiokafka/consumer/group_coordinator.py
joshuaherrera: I'll work on the first solution, updating the private function's signature to accept a `Response` object, since each `JoinGroupResponse` version implements the `Response` [abstract class](https://github.com/joshuaherrera/aiokafka/blob/kip-345/aiokafka/protocol/api.py#L95).

aiokafka/consumer/group_coordinator.py
joshuaherrera: Removed, may be re-added in a new PR,[ see here.](https://github.com/aio-libs/aiokafka/pull/941#discussion_r1405461497)

aiokafka/consumer/group_coordinator.py
```diff
@@ -1202,46 +1217,67 @@ async def perform_group_join(self):
                 metadata = metadata.encode()
             group_protocol = (assignor.name, metadata)
             metadata_list.append(group_protocol)
+            # for KIP-394 we may have to send a second join request
+            try_join = True
+            while try_join:
+                try_join = False
+
+                if self._api_version < (0, 10, 1):
```
tvoinarovskyi, line 1225: nit: Could we abstract out this into a function?

aiokafka/consumer/group_coordinator.py
```diff
@@ -1202,46 +1217,67 @@ async def perform_group_join(self):
                 metadata = metadata.encode()
             group_protocol = (assignor.name, metadata)
             metadata_list.append(group_protocol)
+            # for KIP-394 we may have to send a second join request
```
tvoinarovskyi, line 1220: This sounds like a completely different effort, maybe worth separating out to another PR?

tests/test_consumer.py
```diff
@@ -2052,3 +2052,199 @@ async def on_partitions_assigned(self, assigned):
...
+        # be 2.
+        self.assertEqual(listener1.revoke_mock.call_count, 2)
+        self.assertEqual(listener1.assign_mock.call_count, 2)
+        assert c2_partitions == consumer2.assignment()
+
+        # wait for timeout, consumer1 should get all
+        # partitions after rebalance
+        all_partitions = frozenset(list(c1_partitions) + list(c2_partitions))
+        await consumer2.stop()
+        await asyncio.sleep(25)
```
tvoinarovskyi, line 2153: Could we document here exactly the reason this is 25s? As I understand it is derived from default timeouts, maybe we can make it as low as 10s?
The reason I am asking is that CI has limited time to execute and stops like those increase runtime constantly. A better approach would be to poll for results, for example, wait for every second till listener1.revoke_mock.call_count > 2.

tvoinarovskyi:
Hi [joshuaherrera](https://github.com/joshuaherrera),

Thanks for the PR, will be joining the review. Really nice, clean PR, thanks for the contribution. Added a few small points to consider. 
Could you do a rebase to master, seems like we have a few broken CI runs due to upstream changes?

tests/test_consumer.py
```diff
@@ -2052,3 +2052,199 @@ async def on_partitions_assigned(self, assigned):
...
+        # for both consumers
+        await asyncio.sleep(5)
+        self.assertEqual(listener2.revoke_mock.call_count, 2)
+        self.assertEqual(listener2.assign_mock.call_count, 2)
+        self.assertEqual(listener1.revoke_mock.call_count, 4)
+        self.assertEqual(listener1.assign_mock.call_count, 4)
+
+        # stop consumer2, which will trigger yet another rebalance
+        await consumer2.stop()
+        await asyncio.sleep(15)
```
tvoinarovskyi, line 2248: Same as above

aiokafka/consumer/group_coordinator.py
joshuaherrera: The bulk of this addition is captured in the new `while` loop [wrapping lines 1222-1279](https://github.com/joshuaherrera/aiokafka/blob/kip-345/aiokafka/consumer/group_coordinator.py#L1220-L1279), and the newly added [MemberIdRequired error](https://github.com/joshuaherrera/aiokafka/blob/kip-345/aiokafka/errors.py#L859). [KIP-394](https://cwiki.apache.org/confluence/display/KAFKA/KIP-394%3A+Require+member.id+for+initial+join+group+request) introduces the `MemberIdRequired` error, and according to the KIP, if the client gets this error they should retry the join with the previous member id.

I'd prefer to leave it in since the [original author](https://github.com/robinhood/aiokafka/commit/cd44bdcf57b4aa6f675783ed1043b2d643b0ac54) thought it necessary and the KIP itself calls out that this is "an important complement to KIP-345" but if you think this should be moved to it's own PR (and would delay this PR) I can remove the code related to KIP-394 from this PR.

aiokafka/consumer/group_coordinator.py
joshuaherrera: This depends on the KIP-394 decision, since this is tied to that. If we move that to another PR than this would revert to how we currently handle this in master.

tests/test_consumer.py
joshuaherrera: I'll make the polling change as recommended.

aiokafka/consumer/group_coordinator.py
```diff
@@ -393,15 +400,23 @@ async def _on_join_prepare(self, previous_assignment):
...
             'Invalid assignment protocol: %s' % assignment_strategy
         member_metadata = {}
         all_subscribed_topics = set()
-        for member_id, metadata_bytes in members:
+        for member in members:
+            if isinstance(response, JoinGroupResponse[3]):
+                member_id, group_instance_id, metadata_bytes = member
+            elif (isinstance(response, JoinGroupResponse[0]) or
+                  isinstance(response, JoinGroupResponse[1]) or
+                  isinstance(response, JoinGroupResponse[2])):
```
ods, line 416: _nit:_ 
```suggestion
            elif isinstance(response, (JoinGroupResponse[0], JoinGroupResponse[1], JoinGroupResponse[2])):
```

aiokafka/consumer/group_coordinator.py
```diff
@@ -393,15 +400,23 @@ async def _on_join_prepare(self, previous_assignment):
...
+        assignment_strategy = response.group_protocol
+        members = response.members
         assignor = self._lookup_assignor(assignment_strategy)
         assert assignor, \
             'Invalid assignment protocol: %s' % assignment_strategy
         member_metadata = {}
         all_subscribed_topics = set()
-        for member_id, metadata_bytes in members:
+        for member in members:
+            if isinstance(response, JoinGroupResponse[3]):
```
ods, line 412: _nit:_ It's not obvious that `JoinGroupResponse[3]` is v5, maybe it would be better to use `JoinGroupResponse_vN` explicitly?

joshuaherrera: updated

ods (approved):

aiokafka/consumer/group_coordinator.py
joshuaherrera: @tvoinarovskyi Any updates on next steps? If you or @ods think it's best to remove this and leave it for another PR, I'm happy to do so, since it seems like this is the last unresolved thread for this PR. I don't want to hold this up unnecessarily.

tartieret:
This is very exciting ! Thanks a lot for taking care of this. Do you have any idea when this could be released?

ods:
@tartieret Please [try it out](https://pypi.org/project/aiokafka/0.10.0a0/) and [provide feedback](https://github.com/aio-libs/aiokafka/discussions/951) whether it works as expected.

tartieret:
> @tartieret Please [try it out](https://pypi.org/project/aiokafka/0.10.0a0/) and [provide feedback](https://github.com/aio-libs/aiokafka/discussions/951) whether it works as expected.

I'll have someone in my team look at this over the next few days

igulati:
> @tartieret Please [try it out](https://pypi.org/project/aiokafka/0.10.0a0/) and [provide feedback](https://github.com/aio-libs/aiokafka/discussions/951) whether it works as expected.

hey @ods , replying on behalf of @tartieret, I just tested these changes and they seems to be working for me.

PR #960, opened 2024-01-17
Use cramjam for LZ4

ods:
### Changes

Reduces number of dependencies

### Checklist


vmaurin:
@ods Is there some resource about performance ? We are using lz4 in production on our end, mainly for performance reason and I wonder how cramjam will perform compared to the current lz4 lib

vmaurin:
Replying to myself https://github.com/milesgranger/cramjam/blob/master/cramjam-python/benchmarks/README.md#lz4

ods:
I think looking to cramjam benchmarks is not enough here, as I also switched from level=0 to level=9, because [9 is used by broker by default](https://cwiki.apache.org/confluence/display/KAFKA/KIP-390%3A+Support+Compression+Level). Level is not configurable yet, so this may also affect you.

aisven:
@ods In the context of this change, was the documentation updated?

https://aiokafka.readthedocs.io/en/stable/

It still talks about installing optional dependencies like `pip install 'aiokafka[lz4]'`.

Not sure if the documentation needs more updating?

ods:
Hi @aisven, the documentation is correct: the name of extra wasn't changed, but now it installs another dependency for LZ4 support. The idea is that aiokafka users normally shouldn't care about implementation details like what dependency is used to implement particular codec.

PR #969, opened 2024-01-25
Implement KIP-204 : DeleteRecords API

### LINKED ISSUE #967: Add delete_records to the admin client

We should have a method on the admin client to delete records : API 21 https://kafka.apache.org/protocol#The_Messages_DeleteRecords

It is useful when doing stream processing and have some logic to cleanup repartitions/shuffle topics

**Additional context**

It was added a while ago to the java client, see https://cwiki.apache.org/confluence/display/KAFKA/KIP-204+%3A+Adding+records+deletion+operation+to+the+new+Admin+Client+API

vmaurin:
When doing stream processing, it is convinient to use "transient" topic :
* retention time is infinite
* records get deleted when consumed

The java kafka streams client is using the deleteRecords of the admin client to perform this operation. It is lacking in aiokafka

The KIP reference https://cwiki.apache.org/confluence/display/KAFKA/KIP-204+%3A+Adding+records+deletion+operation+to+the+new+Admin+Client+API

refs #967

### Changes

Fixes #967 

### Checklist

### DIFF
--- a/aiokafka/admin/client.py
+++ b/aiokafka/admin/client.py
@@ -626,12 +626,7 @@ class AIOKafkaAdminClient:
         """
         version = self._matching_api_version(DeleteRecordsRequest)
 
-        if self._version_info[MetadataRequest[0].API_KEY] < (0, 10):
-            metadata_request = MetadataRequest[0]([])
-        else:
-            metadata_request = MetadataRequest[1](None)
-
-        metadata = await self._send_request(metadata_request)
+        metadata = await self._get_cluster_metadata()
 
         self._client.cluster.update_metadata(metadata)
 
--- a/aiokafka/admin/client.py
+++ b/aiokafka/admin/client.py
@@ -646,8 +646,7 @@ class AIOKafkaAdminClient:
         for leader, delete_request in requests.items():
             request = req_cls(
                 self._convert_records_to_delete(delete_request),
-                timeout_ms or self._request_timeout_ms,
-                {},
+                timeout_ms or self._request_timeout_ms
             )
             response = await self._client.send(leader, request)
             for topic, partitions in response.topics:
@@ -665,8 +664,7 @@ class AIOKafkaAdminClient:
         return [
             (
                 topic,
-                [(partition, rec.before_offset, {}) for partition, rec in records],
-                {},
+                [(partition, rec.before_offset) for partition, rec in records]
             )
             for topic, records in records_to_delete.items()
         ]
--- a/aiokafka/admin/client.py
+++ b/aiokafka/admin/client.py
@@ -646,7 +646,7 @@ class AIOKafkaAdminClient:
         for leader, delete_request in requests.items():
             request = req_cls(
                 self._convert_records_to_delete(delete_request),
-                timeout_ms or self._request_timeout_ms
+                timeout_ms or self._request_timeout_ms,
             )
             response = await self._client.send(leader, request)
             for topic, partitions in response.topics:
@@ -662,9 +662,6 @@ class AIOKafkaAdminClient:
         records_to_delete: Dict[str, List[Tuple[int, RecordsToDelete]]],
     ):
         return [
-            (
-                topic,
-                [(partition, rec.before_offset) for partition, rec in records]
-            )
+            (topic, [(partition, rec.before_offset) for partition, rec in records])
             for topic, records in records_to_delete.items()
         ]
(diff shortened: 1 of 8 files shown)


aiokafka/admin/client.py
```diff
@@ -605,3 +612,66 @@ async def list_consumer_group_offsets(
...
+        :return: Appropriate version of DeleteRecordsResponse class.
+        """
+        version = self._matching_api_version(DeleteRecordsRequest)
+
+        if self._version_info[MetadataRequest[0].API_KEY] < (0, 10):
+            metadata_request = MetadataRequest[0]([])
+        else:
+            metadata_request = MetadataRequest[1](None)
+
+        metadata = await self._send_request(metadata_request)
```
ods, line 634:
```suggestion
        metadata = await self._get_cluster_metadata(topics=None)
```

vmaurin: With higher metadata protocol version `self._client.cluster.update_metadata(metadata)` doesn't work in the current version

```
            if error_type is Errors.NoError:
                _new_partitions[topic] = {}
>               for p_error, partition, leader, replicas, isr in partitions:
E               ValueError: too many values to unpack (expected 5)
```

Should I rather try to fix `update_metadata` or use the existing snippet to restrict ourselves to v0 or v1 ?

vmaurin: I did the fix (that should be backward compatible)

aiokafka/admin/client.py
```diff
@@ -605,3 +612,66 @@ async def list_consumer_group_offsets(
...
+
+        self._client.cluster.update_metadata(metadata)
+
+        requests = defaultdict(lambda: defaultdict(list))
+        responses = {}
+
+        for tp, records in records_to_delete.items():
+            leader = self._client.cluster.leader_for_partition(tp)
```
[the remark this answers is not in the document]
vmaurin: Sure, but here as I understood the KIP and the java client, we should aim the query to the leader node, while `create_topics`, `delete_topics`, `create_partitions` should target the controller broker. As this PR doesn't work yet, maybe it is better to do a separated PR for it ? (I can eventually give it a shot)

aiokafka/admin/client.py
ods: I think it would be better to fix it. As I understand, it could be as simple as
```python
for p_error, partition, leader, replicas, isr, *_ in partitions:
```
\+ a comment with explanation. Or do you see some reason against this?

vmaurin: I just did the same fix, but without the comment, let me add one

aiokafka/admin/client.py
```diff
@@ -605,3 +612,61 @@ async def list_consumer_group_offsets(
...
+
+    @staticmethod
+    def _convert_records_to_delete(
+        records_to_delete: Dict[str, List[Tuple[int, RecordsToDelete]]],
+    ):
+        return [
+            (
+                topic,
+                [(partition, rec.before_offset, {}) for partition, rec in records],
+                {},
```
ods, line 669: Tags are available in v2 only.

aiokafka/protocol/admin.py
```diff
@@ -1276,3 +1276,124 @@ class ListPartitionReassignmentsRequest_v0(Request):
...
+    DeleteRecordsRequest_v0,
+    DeleteRecordsRequest_v1,
+    DeleteRecordsRequest_v2,
+]
+
+DeleteRecordsResponse = [
+    DeleteRecordsResponse_v0,
+    DeleteRecordsResponse_v1,
+    DeleteRecordsResponse_v2,
+]
```
ods, line 1399: A possible way to go for now:
```suggestion
DeleteRecordsRequest = [
    DeleteRecordsRequest_v0,
    DeleteRecordsRequest_v1,
    # FIXME: We have some problems with `TaggedFields`
    # DeleteRecordsRequest_v2,
]

DeleteRecordsResponse = [
    DeleteRecordsResponse_v0,
    DeleteRecordsResponse_v1,
    # FIXME: We have some problems with `TaggedFields`
    # DeleteRecordsResponse_v2,
]
```

vmaurin: Updated

aiokafka/admin/client.py
vmaurin: Removed

ods:
@vmaurin  I think it's ready to merge, no objections?

PR #978, opened 2024-02-11
Fix type annotation for `AIOKafkaAdminClient.create_partitions`

### LINKED ISSUE #977: invalid Type AioKafkaAdminClient  create_partitions
**Describe the bug**
there is a type error or wrong type

 - aiokafka version : `v0.10.0`
 - Kafka Broker version : `7.1.2-ccs (Commit:99a40ee4675ee147)`

**Reproducible example**
in `admin/client.py`
```python
    @staticmethod
    def _convert_topic_partitions(topic_partitions: Dict[str, TopicPartition]):
        return [(topic_name, (new_part.total_count, new_part.new_assignments))
                for topic_name, new_part in topic_partitions.items()]
```
while in `struct.py` `TopicPartition` defined like this 
```python
class TopicPartition(NamedTuple):
    """A topic and partition tuple"""

    topic: str
    "A topic name"

    partition: int
    "A partition id"
```
and there is no `total_count` and `new_assignments` in this type 
it must be `admin/new_partitions.py`
```python
class NewPartitions:
    """A class for new partition creation on existing topics.  Note that the
    length of new_assignments, if specified, must be the difference between the
    new total number of partitions and the existing number of partitions.
    Arguments:
        total_count (int):
            the total number of partitions that should exist on the topic
        new_assignments ([[int]]):
            an array of arrays of replica assignments for new partitions.
            If not set, broker assigns replicas per an internal algorithm.
    """

    def __init__(
        self,
        total_count,
        new_assignments=None
    ):
        self.total_count = total_count
        self.new_assignments = new_assignments

```

alm0ra:
Fixes #977 

### Changes

Fixes wrong typo in create_partitions() for the admin client

### Checklist

### DIFF
--- a/aiokafka/admin/client.py
+++ b/aiokafka/admin/client.py
@@ -27,6 +27,7 @@ from aiokafka.protocol.api import Request, Response
 from aiokafka.protocol.commit import GroupCoordinatorRequest, OffsetFetchRequest
 from aiokafka.protocol.metadata import MetadataRequest
 from aiokafka.structs import OffsetAndMetadata, TopicPartition
+from aiokafka.admin import NewPartitions
 
 from .config_resource import ConfigResource, ConfigResourceType
 from .new_topic import NewTopic
@@ -395,7 +396,7 @@ class AIOKafkaAdminClient:
         return broker_resources, topic_resources
 
     @staticmethod
-    def _convert_topic_partitions(topic_partitions: Dict[str, TopicPartition]):
+    def _convert_topic_partitions(topic_partitions: Dict[str, NewPartitions]):
         return [
             (topic_name, (new_part.total_count, new_part.new_assignments))
             for topic_name, new_part in topic_partitions.items()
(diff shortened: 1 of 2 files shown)


alm0ra:
@tvoinarovskyi @ods 

there is no documentation provided for `AIOKafkaAdminClient` so there is nothing to update and changes 
also for test, there is a test for this part [test_admin](https://github.com/aio-libs/aiokafka/blob/master/tests/test_admin.py) that satisfy this change

aiokafka/admin/client.py
```diff
@@ -27,6 +27,7 @@
 from aiokafka.protocol.commit import GroupCoordinatorRequest, OffsetFetchRequest
 from aiokafka.protocol.metadata import MetadataRequest
 from aiokafka.structs import OffsetAndMetadata, TopicPartition
+from aiokafka.admin import NewPartitions
```
ods, line 30: It would be better to avoid importing from parent package
```suggestion
from .new_partitions import NewPartitions
```

alm0ra: @ods Already fixed and sort imports

alm0ra:
@ods All resolved i think it's ready

Issue #980, opened 2024-02-13
Proposal to Add Type Hints 


alm0ra:
**Describe the solution you'd like**
I'd like to propose adding type hints to the `Aiokafka` project. I've used `mypy` to analyze the codebase, and I believe that adding type hints could significantly improve the code's readability, maintainability, and robustness.

result for mypy
``` shell
Found 2475 errors in 109 files (checked 133 source files)
```
By adding type hints, developers can better understand function signatures and catch potential bugs early in the development process. Additionally, it can make it easier for newcomers to contribute to the project and for existing contributors to navigate the codebase.

I'm willing to contribute to this effort if the maintainers are interested in pursuing it. Adding type hints could be a valuable enhancement to the Aiokafka project.

I can start with [aiokafka/consumer/consumer.py](https://github.com/aio-libs/aiokafka/tree/master/aiokafka/consumer) and fix all issue in this file first and continue with other files in another PR's or you can assign others to contribute

alm0ra:
@ods

ods:
Yes, we're interested in adding type hints. I'd like to highlight some issues here.

- Incorrect types are often worse that missing types
- Gradual approach doesn't work good here: you never sure types are correct and consistent, until everything is annotated
- Some features like custom serialization functions don't allow us to provide good annotations. So, we may want to consider deprecating such features first.

alm0ra:
Thank you for bringing up those concerns. I understand the importance of ensuring that type hints are accurate and consistent. Regarding the "Gradual approach doesn't work good here" point, I believe we can still make progress by prioritizing certain areas for annotation while acknowledging that some parts may remain untyped initially.

By leveraging tools like MyPy, we can focus on annotating sections of the codebase where types are clearer, thereby incrementally improving the overall typing coverage. We can then gradually address the more complex or ambiguous sections, such as the custom serialization functions, as we gain more clarity on their typings.

alm0ra:
#981 should I close this PR?

ods:
> https://github.com/aio-libs/aiokafka/pull/981 should I close this PR?

I hope to find time on weekend to look through it.

antonagestam:
Beartype could be of interest here. I've successfully used it in unit tests to increase confidence that type hints are accurate. It can be injected by using an early stage pytest hook.

https://github.com/beartype/beartype

ods:
[...]

PR #993, opened 2024-03-24
Fix typing bugs found by beartype

ods:
Try [pytest-beartype](https://github.com/beartype/pytest-beartype), as [suggested](https://github.com/aio-libs/aiokafka/issues/980#issuecomment-1987811336) by @antonagestam . Only two bugs found so far, doesn't seem enough to add it to CI.

### DIFF
--- a/aiokafka/admin/client.py
+++ b/aiokafka/admin/client.py
@@ -2,7 +2,7 @@ import asyncio
 import logging
 from collections import defaultdict
 from ssl import SSLContext
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union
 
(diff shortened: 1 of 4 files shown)


antonagestam:
Cool stuff, I hope it at least brings some confidence to the type hints 👍

ods:
> I hope it at least brings some confidence to the type hints 👍

Unfortunately it doesn't catch even simple and obvious bugs. So, I'd like to find another runtime checker.

antonagestam:
@ods FWIW, there's also typeguard: https://github.com/agronholm/typeguard

ods:
> FWIW, there's also typeguard: https://github.com/agronholm/typeguard

Right, but the use is not as simple as "install and add command line option", as it is with beartype

PR #999, opened 2024-04-13
add typing to aiokafka/protocol/*

### LINKED ISSUE #980: Proposal to Add Type Hints

dimastbk:
### Changes

partial fixes #980

Add typing to all modules in protocol package.

### Checklist

### DIFF
--- a/aiokafka/protocol/admin.py
+++ b/aiokafka/protocol/admin.py
@@ -1,3 +1,6 @@
+from collections.abc import Iterable
+from typing import Dict, Optional, Tuple
+
 from .api import Request, Response
 from .types import (
     Array,
@@ -68,16 +71,16 @@ class ApiVersionRequest_v2(Request):
     SCHEMA = ApiVersionRequest_v0.SCHEMA
 
 
-ApiVersionRequest = [
+ApiVersionRequest = (
     ApiVersionRequest_v0,
     ApiVersionRequest_v1,
     ApiVersionRequest_v2,
-]
-ApiVersionResponse = [
+)
+ApiVersionResponse = (
     ApiVersionResponse_v0,
     ApiVersionResponse_v1,
     ApiVersionResponse_v2,
-]
+)
 
 
 class CreateTopicsResponse_v0(Response):
@@ -196,18 +199,18 @@ class CreateTopicsRequest_v3(Request):
     SCHEMA = CreateTopicsRequest_v1.SCHEMA
 
 
-CreateTopicsRequest = [
+CreateTopicsRequest = (
     CreateTopicsRequest_v0,
     CreateTopicsRequest_v1,
     CreateTopicsRequest_v2,
     CreateTopicsRequest_v3,
-]
-CreateTopicsResponse = [
+)
+CreateTopicsResponse = (
     CreateTopicsResponse_v0,
     CreateTopicsResponse_v1,
     CreateTopicsResponse_v2,
     CreateTopicsResponse_v3,
-]
+)
 
 
 class DeleteTopicsResponse_v0(Response):
@@ -267,18 +270,18 @@ class DeleteTopicsRequest_v3(Request):
     SCHEMA = DeleteTopicsRequest_v0.SCHEMA
 
 
-DeleteTopicsRequest = [
+DeleteTopicsRequest = (
     DeleteTopicsRequest_v0,
     DeleteTopicsRequest_v1,
     DeleteTopicsRequest_v2,
     DeleteTopicsRequest_v3,
-]
-DeleteTopicsResponse = [
+)
+DeleteTopicsResponse = (
     DeleteTopicsResponse_v0,
     DeleteTopicsResponse_v1,
     DeleteTopicsResponse_v2,
     DeleteTopicsResponse_v3,
-]
+)
 
 
 class ListGroupsResponse_v0(Response):
@@ -333,16 +336,16 @@ class ListGroupsRequest_v2(Request):
     SCHEMA = ListGroupsRequest_v0.SCHEMA
 
 
-ListGroupsRequest = [
+ListGroupsRequest = (
     ListGroupsRequest_v0,
     ListGroupsRequest_v1,
     ListGroupsRequest_v2,
-]
-ListGroupsResponse = [
+)
+ListGroupsResponse = (
     ListGroupsResponse_v0,
     ListGroupsResponse_v1,
     ListGroupsResponse_v2,
-]
+)
 
 
 class DescribeGroupsResponse_v0(Response):
@@ -429,8 +432,8 @@ class DescribeGroupsResponse_v3(Response):
                         ("member_assignment", Bytes),
                     ),
                 ),
+                ("authorized_operations", Int32),
             ),
-            ("authorized_operations", Int32),
         ),
     )
 
@@ -465,18 +468,18 @@ class DescribeGroupsRequest_v3(Request):
     )
 
 
-DescribeGroupsRequest = [
+DescribeGroupsRequest = (
     DescribeGroupsRequest_v0,
     DescribeGroupsRequest_v1,
     DescribeGroupsRequest_v2,
     DescribeGroupsRequest_v3,
-]
-DescribeGroupsResponse = [
+)
+DescribeGroupsResponse = (
     DescribeGroupsResponse_v0,
     DescribeGroupsResponse_v1,
     DescribeGroupsResponse_v2,
     DescribeGroupsResponse_v3,
-]
+)
 
 
 class SaslHandShakeResponse_v0(Response):
@@ -507,8 +510,8 @@ class SaslHandShakeRequest_v1(Request):
     SCHEMA = SaslHandShakeRequest_v0.SCHEMA
 
 
-SaslHandShakeRequest = [SaslHandShakeRequest_v0, SaslHandShakeRequest_v1]
-SaslHandShakeResponse = [SaslHandShakeResponse_v0, SaslHandShakeResponse_v1]
+SaslHandShakeRequest = (SaslHandShakeRequest_v0, SaslHandShakeRequest_v1)
+SaslHandShakeResponse = (SaslHandShakeResponse_v0, SaslHandShakeResponse_v1)
 
 
 class DescribeAclsResponse_v0(Response):
@@ -610,8 +613,8 @@ class DescribeAclsRequest_v2(Request):
     SCHEMA = DescribeAclsRequest_v1.SCHEMA
 
 
-DescribeAclsRequest = [DescribeAclsRequest_v0, DescribeAclsRequest_v1]
-DescribeAclsResponse = [DescribeAclsResponse_v0, DescribeAclsResponse_v1]
+DescribeAclsRequest = (DescribeAclsRequest_v0, DescribeAclsRequest_v1)
+DescribeAclsResponse = (DescribeAclsResponse_v0, DescribeAclsResponse_v1)
 
 
 class CreateAclsResponse_v0(Response):
@@ -671,8 +674,8 @@ class CreateAclsRequest_v1(Request):
     )
 
 
-CreateAclsRequest = [CreateAclsRequest_v0, CreateAclsRequest_v1]
-CreateAclsResponse = [CreateAclsResponse_v0, CreateAclsResponse_v1]
+CreateAclsRequest = (CreateAclsRequest_v0, CreateAclsRequest_v1)
+CreateAclsResponse = (CreateAclsResponse_v0, CreateAclsResponse_v1)
 
 
 class DeleteAclsResponse_v0(Response):
@@ -771,8 +774,8 @@ class DeleteAclsRequest_v1(Request):
     )
 
 
-DeleteAclsRequest = [DeleteAclsRequest_v0, DeleteAclsRequest_v1]
-DeleteAclsResponse = [DeleteAclsResponse_v0, DeleteAclsResponse_v1]
+DeleteAclsRequest = (DeleteAclsRequest_v0, DeleteAclsRequest_v1)
+DeleteAclsResponse = (DeleteAclsResponse_v0, DeleteAclsResponse_v1)
 
 
 class AlterConfigsResponse_v0(Response):
@@ -828,8 +831,8 @@ class AlterConfigsRequest_v1(Request):
     SCHEMA = AlterConfigsRequest_v0.SCHEMA
 
 
-AlterConfigsRequest = [AlterConfigsRequest_v0, AlterConfigsRequest_v1]
-AlterConfigsResponse = [AlterConfigsResponse_v0, AlterConfigsRequest_v1]
+AlterConfigsRequest = (AlterConfigsRequest_v0, AlterConfigsRequest_v1)
+AlterConfigsResponse = (AlterConfigsResponse_v0, AlterConfigsRequest_v1)
 
 
 class DescribeConfigsResponse_v0(Response):
@@ -969,16 +972,16 @@ class DescribeConfigsRequest_v2(Request):
     SCHEMA = DescribeConfigsRequest_v1.SCHEMA
 
 
-DescribeConfigsRequest = [
+DescribeConfigsRequest = (
     DescribeConfigsRequest_v0,
     DescribeConfigsRequest_v1,
     DescribeConfigsRequest_v2,
-]
-DescribeConfigsResponse = [
+)
+DescribeConfigsResponse = (
     DescribeConfigsResponse_v0,
     DescribeConfigsResponse_v1,
     DescribeConfigsResponse_v2,
-]
+)
 
 
 class SaslAuthenticateResponse_v0(Response):
@@ -1016,14 +1019,14 @@ class SaslAuthenticateRequest_v1(Request):
     SCHEMA = SaslAuthenticateRequest_v0.SCHEMA
 
 
-SaslAuthenticateRequest = [
+SaslAuthenticateRequest = (
     SaslAuthenticateRequest_v0,
     SaslAuthenticateRequest_v1,
-]
-SaslAuthenticateResponse = [
+)
+SaslAuthenticateResponse = (
     SaslAuthenticateResponse_v0,
     SaslAuthenticateResponse_v1,
-]
+)
 
 
 class CreatePartitionsResponse_v0(Response):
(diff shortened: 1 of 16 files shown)


aiokafka/protocol/admin.py
```diff
@@ -68,16 +71,16 @@ class ApiVersionRequest_v2(Request):
     SCHEMA = ApiVersionRequest_v0.SCHEMA
 
 
-ApiVersionRequest = [
+ApiVersionRequest = (
```
dimastbk, line 74: For better type inference in mypy and pyright.

aiokafka/protocol/types.py
```diff
@@ -1,10 +1,31 @@
...
+)
+
+from typing_extensions import Buffer, TypeAlias
 
 from .abstract import AbstractType
 
+T = TypeVar("T")
 
-def _pack(f, value):
+ValueT: TypeAlias = Union[Type[AbstractType[Any]], "String", "Array", "Schema"]
```
ods, line 25: Interesting trick. But it would be better to explicitly declare their common interface.

aiokafka/protocol/metadata.py
```diff
@@ -238,23 +242,23 @@ class MetadataRequest_v5(Request):
...
     RESPONSE_TYPE = MetadataResponse_v5
     SCHEMA = MetadataRequest_v4.SCHEMA
-    ALL_TOPICS = -1  # Null Array (len -1) for topics returns all topics
-    NO_TOPICS = None  # Empty array (len 0) for topics returns no topics
+    ALL_TOPICS: Optional[int] = -1  # Null Array (len -1) for topics returns all topics
+    NO_TOPICS: None = None  # Empty array (len 0) for topics returns no topics
 
 
-MetadataRequest = [
+MetadataRequest = (
```
ods, line 249: Do I understand correctly that the aim of this change is to solve problem with variance? Semantically tuple doesn't fit well here.

aiokafka/protocol/produce.py
```diff
@@ -190,20 +183,27 @@ class ProduceRequest_v0(ProduceRequest):
         ),
     )
 
+    required_acks: int
 
-class ProduceRequest_v1(ProduceRequest):
+    def expect_response(self) -> bool:
```
ods, line 188: What is the reason for moving it from base class?

aiokafka/protocol/api.py
```diff
@@ -48,32 +58,46 @@ class ResponseHeader_v1(Struct):
...
-    @abc.abstractproperty
-    def API_KEY(self):
+    @property
+    @abc.abstractmethod
+    def API_KEY(self) -> int:  # pyright:ignore[reportRedeclaration]
         """Integer identifier for api request"""
 
-    @abc.abstractproperty
-    def API_VERSION(self):
+    API_KEY: ClassVar[int]  # type: ignore[no-redef]
```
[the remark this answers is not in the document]
ods: Let's leave only second definition instead of adding `type: ignore` comments

aiokafka/protocol/types.py
ods: Don't mind, I think this solution is good enough for now to go farther with typing. But the target (somewhen in the future) it would be nice to get typing for all parsed messages, e.g. with pydantic-style schemes.

aiokafka/protocol/metadata.py
dimastbk: I think the first variant is more correct.

```python
MetadataRequestTuple = (
    MetadataRequest_v0,
    MetadataRequest_v1,
    MetadataRequest_v2,
    MetadataRequest_v3,
    MetadataRequest_v4,
    MetadataRequest_v5,
)

reveal_type(MetadataRequestTuple)  # Revealed type is "Tuple[def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v0, def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v1, def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v2, def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v3, def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v4, def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v5]"
reveal_type(MetadataRequestTuple[0])  # Revealed type is "def (*args: Any, **kwargs: Any) -> aiokafka.protocol.metadata.MetadataRequest_v0"

MetadataRequestList = [
    MetadataRequest_v0,
    MetadataRequest_v1,
    MetadataRequest_v2,
    MetadataRequest_v3,
    MetadataRequest_v4,
    MetadataRequest_v5,
]

reveal_type(MetadataRequestList)  # Revealed type is "builtins.list[def (*args: Any, **kwargs: Any) -> aiokafka.protocol.api.Request]"
reveal_type(MetadataRequestList[0])  # Revealed type is "def (*args: Any, **kwargs: Any) -> aiokafka.protocol.api.Request"
```

aiokafka/protocol/produce.py
dimastbk: Because the base class doesn't contain `required_acks` in the schema.

aiokafka/protocol/api.py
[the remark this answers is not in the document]
dimastbk: But without abstract property mypy won't check the required classvar.

dimastbk:
Maybe rewrite all schemas using dataclasses? If it is not a problem for performance.

aiokafka/protocol/types.py
```diff
@@ -253,36 +300,36 @@ def encode(cls, value):
...
+            if not (b & 0x80):  # type: ignore[operator]
                 break
-            value |= (b & 0x7F) << i
+            value |= (b & 0x7F) << i  # type: ignore[operator]
             i += 7
             if i > 63:
                 raise ValueError(f"Invalid value {value}")
-        value |= b << i
+        value |= b << i  # type: ignore[operator]
         return (value >> 1) ^ -(value & 1)
```
dimastbk, line 329: Smth strange with types in this class, but I didn't find using VarInt64 in aiokafka, so I don't know what do with these errors.

aiokafka/protocol/produce.py
ods: Do I understand correctly, that all subclasses of `ProduceRequestBase` have `required_acks`? If so, both `required_acks` and `expect_response` redefinition can be places there.

aiokafka/protocol/api.py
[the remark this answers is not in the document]
ods: The current code is definitely incorrect (those definitions contradict each other), mypy is right when complaining about redefinition. `abstractproperty` is deprecated and should be replaced. So, I see 2 options here: either to use `@classmethod` + `@property`, or to declare is as `ClassVar`. I see no problems about pyright complaints as it's not used in CI here (and won't be used soon, as it has a lot of other problems) and protocol classes don't show up in public interface of aiokafka.

aiokafka/protocol/metadata.py
ods: The semantic of tuple is: a fixed number of fields each having it's own type, i.e. like struct, but with access by index instead of name. The semantic of list is: a variable array of items with the same type. Here, we have an array of items each being a request type.

aiokafka/protocol/types.py
ods: The problem here is that `data.read(1)` returns `bytes` and you need to unpack it the same way as in other places to get `int`. Another option might be to use `data.read(1)[0]`, but let's keep it consistent.

dimastbk: Thanks, fixed

aiokafka/protocol/produce.py
dimastbk: Ok, done

aiokafka/protocol/api.py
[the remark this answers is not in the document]
dimastbk: Ok, I removed ClassVar

aiokafka/protocol/metadata.py
dimastbk: But in some cases we need to know type of "field", for example (aiokafka/client.py):

```python
        if self.api_version > (0, 11):
            request = FindCoordinatorRequest[1](coordinator_key, coordinator_type)
        else:
            # Group coordination only
            assert (
                coordinator_type == CoordinationType.GROUP
            ), "No transactions for older brokers"
            request = FindCoordinatorRequest[0](coordinator_key)
```

ods: Do you see any reason of using `FindCoordinatorRequest[1]` instead of `FindCoordinatorRequest_v1` there? To me the variant with index is somewhat less reliable way to refer to particular version of request.

dimastbk: You are right, but it's just example from aiokafka :)

Ok, I reverted my changes

aiokafka/protocol/message.py
```diff
@@ -77,66 +114,85 @@ def timestamp_type(self):
...
         # Partial decode required to determine message version
-        base_fields = cls.SCHEMAS[0].fields[0:3]
+        base_fields = cast(
+            Tuple[Type[UInt32], Type[Int8], Type[Int8]], cls.SCHEMAS[0].fields[0:3]
+        )
         crc, magic, attributes = (field.decode(data) for field in base_fields)
         remaining = cls.SCHEMAS[magic].fields[3:]
-        fields = [field.decode(data) for field in remaining]
+        fields = tuple(field.decode(data) for field in remaining)
+        magic = cast(Literal[0, 1], magic)
```
ods, line 155: It's incorrect to narrow the type _before_ verification

aiokafka/protocol/message.py
```diff
@@ -77,66 +114,85 @@ def timestamp_type(self):
...
+    def decode(cls, data: Union[io.BytesIO, bytes]) -> Self:
+        _validated_crc: Optional[int] = None
         if isinstance(data, bytes):
             _validated_crc = crc32(data[4:])
             data = io.BytesIO(data)
         # Partial decode required to determine message version
-        base_fields = cls.SCHEMAS[0].fields[0:3]
+        base_fields = cast(
+            Tuple[Type[UInt32], Type[Int8], Type[Int8]], cls.SCHEMAS[0].fields[0:3]
+        )
```
ods, line 151: It's strange to duplicate scheme here and cast simultaneously. Could we describe these common fields in separate class variable is a way, that allows mypy to deduct the type of each field?

aiokafka/protocol/message.py
```diff
@@ -233,10 +298,18 @@ def decode(cls, data, bytes_to_read=None):
         return items
 
     @classmethod
-    def repr(cls, messages):
+    def repr(
+        cls,
+        messages: Union[
+            io.BytesIO,
```
ods, line 304: Why `io.BytesIO` is here?

aiokafka/protocol/metadata.py
```diff
@@ -187,34 +189,36 @@ class MetadataRequest_v0(Request):
...
 
 class MetadataRequest_v1(Request):
     API_KEY = 3
     API_VERSION = 1
     RESPONSE_TYPE = MetadataResponse_v1
     SCHEMA = MetadataRequest_v0.SCHEMA
-    ALL_TOPICS = -1  # Null Array (len -1) for topics returns all topics
-    NO_TOPICS = None  # Empty array (len 0) for topics returns no topics
+    ALL_TOPICS: Optional[int] = -1  # Null Array (len -1) for topics returns all topics
+    NO_TOPICS: None = None  # Empty array (len 0) for topics returns no topics
```
ods, line 203: I don't see any use of these 2 fields. Should we consider removing them? On the other hand comments to them contain important information.

aiokafka/protocol/message.py
```diff
@@ -47,7 +51,40 @@ class Message(Struct):
...
+    ) -> None: ...
+
+    def __init__(
+        self,
+        *,
+        value: Optional[bytes],
+        key: Optional[bytes] = None,
+        magic: Literal[0, 1] = 0,
+        attributes: int = 0,
+        crc: int = 0,
```
ods, line 85: Do we need default values for them? I'm especially uncomfortable seeing it for `crc`

aiokafka/protocol/message.py
dimastbk: Because `if isinstance(messages, io.BytesIO):` three lines below

aiokafka/protocol/message.py
```diff
@@ -47,7 +51,39 @@ class Message(Struct):
         22  # crc(4), magic(1), attributes(1), timestamp(8), key+value size(4*2)
     )
 
-    def __init__(self, value, key=None, magic=0, attributes=0, crc=0, timestamp=None):
+    @overload
+    def __init__(
+        self,
+        *,
```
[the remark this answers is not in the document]
dimastbk: timestamp always None if magic == 0

aiokafka/protocol/message.py
dimastbk: I think no. I can remove all default values, except timestamp.

aiokafka/protocol/message.py
dimastbk: I don't know, how I can do this without 1) generic or 2) mypy plugin.

aiokafka/protocol/struct.py
```diff
@@ -23,27 +25,29 @@ def __init__(self, *args, **kwargs):
...
+    def __repr__(self) -> str:
+        key_vals: List[str] = []
         for name, field in zip(self.SCHEMA.names, self.SCHEMA.fields):
             key_vals.append(f"{name}={field.repr(self.__dict__[name])}")
         return self.__class__.__name__ + "(" + ", ".join(key_vals) + ")"
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Struct):
+            return False
```
dimastbk, line 50: Maybe return NotImplemented?

aiokafka/protocol/metadata.py
dimastbk: Moved to docstring

aiokafka/protocol/message.py
ods: Looks like it's never used. It might be useful with `bytes` here (which is actually won't work). But "list of bytes stream" as input looks very strange to me. Could you imagine a case where we have a list of streams? Let's just drop this branch.

aiokafka/protocol/message.py
```diff
@@ -1,6 +1,10 @@
 import io
 import time
 from binascii import crc32
+from collections.abc import Iterable
```
ods, line 4:
```suggestion
from typing import Iterable
```
It doesn't work in 3.8, which we have to support yet:
```python
>>> from collections.abc import Iterable
>>> Iterable[int]
  File "<stdin>", line 1, in <module>
TypeError: 'ABCMeta' object is not subscriptable
>>> from typing import Iterable
>>> Iterable[int]
typing.Iterable[int]
```

aiokafka/protocol/types.py
```diff
@@ -1,10 +1,31 @@
 import struct
+from collections.abc import Sequence
```
ods, line 2:
```suggestion
from typing import Sequence
```

ods (changes requested):
Hi Dmitry!
Thank you very much, you did a huge piece of work! And sorry for delay with review. I hope we'll finish it very soon.

aiokafka/protocol/message.py
```diff
@@ -167,21 +223,25 @@
 
         return MessageSet.decode(raw_bytes, bytes_to_read=len(raw_bytes))
 
-    def __hash__(self):
+    def __hash__(self) -> int:
```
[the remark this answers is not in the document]
ods: BTW, this has 2 options for the schema, so the `__eq__` (as `__repr__` too) inherited from `Struct` is not correct here. Could you please add a `FIXME` comment to fix them?

aiokafka/protocol/message.py
```diff
@@ -77,66 +113,86 @@ def timestamp_type(self):
...
+            magic = cast(Literal[1], magic)
+            fields = cast(Tuple[int, Optional[bytes], Optional[bytes]], fields)
+            msg = cls(
+                value=fields[-1],
+                key=fields[-2],
+                magic=magic,
+                attributes=attributes,
+                crc=crc,
+                timestamp=fields[0],
+            )
```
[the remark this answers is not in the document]
dimastbk: Done. But without loop over fields because of https://github.com/python/mypy/issues/9230.

aiokafka/protocol/message.py
dimastbk: Done. But without loop over fields because of https://github.com/python/mypy/issues/9230.

aiokafka/protocol/message.py
dimastbk: Why "list of bytes stream", if it is just BytesIO or list of messages?

ods: `BytesIO` = stream

aiokafka/protocol/message.py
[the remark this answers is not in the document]
ods: Interesting. Actually I tried this approach for `BASE_FIELDS` and mypy didn't complain. Ok, let's keep it this way for now.

aiokafka/protocol/struct.py
ods: Where can it make difference? When comparing with `mock.ANY`? Ok, let's return `NotImplemented`

ods (approved):

ods:
Hmm, we forgot about `test_protocol.py` and `test_protocol_object_conversion.py`. They are important, because they allow us to verify how types in `protocol` module actually work. Unfortunately, they don't. We need to do something with attributes from schema, otherwise everything will in `# type: ignore`.

PR #1005, opened 2024-05-06
add typing to tests/test_protocol*

dimastbk:
### Changes

Fixes #

### Checklist

### DIFF
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -285,8 +285,9 @@ def test_decode_fetch_response_partial() -> None:
         ]
     )
     resp = FetchResponse_v0.decode(io.BytesIO(encoded))
-    assert len(resp.topics) == 1  # type: ignore[attr-defined]
-    topic, partitions = resp.topics[0]  # type: ignore[attr-defined]
+    assert resp.topics is not None
+    assert len(resp.topics) == 1
+    topic, partitions = resp.topics[0]
     assert topic == "foobar"
     assert len(partitions) == 2
 
@@ -303,7 +304,7 @@ def test_struct_unrecognized_kwargs() -> None:
 
 def test_struct_missing_kwargs() -> None:
     fr = FetchRequest_v0(max_wait_time=100)
-    assert fr.min_bytes is None  # type: ignore[attr-defined]
+    assert fr.min_bytes is None
 
 
 def test_unsigned_varint_serde() -> None:
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -1,13 +1,14 @@
 import io
 import struct
+from typing import Type
 
 import pytest
 
 from aiokafka.protocol.api import Request, RequestHeader_v0, Response
-from aiokafka.protocol.commit import GroupCoordinatorRequest
-from aiokafka.protocol.fetch import FetchRequest, FetchResponse
+from aiokafka.protocol.commit import GroupCoordinatorRequest_v0
+from aiokafka.protocol.fetch import FetchRequest_v0, FetchResponse_v0
 from aiokafka.protocol.message import Message, MessageSet, PartialMessage
-from aiokafka.protocol.metadata import MetadataRequest
+from aiokafka.protocol.metadata import MetadataRequest_v0
 from aiokafka.protocol.types import (
     CompactArray,
     CompactBytes,
@@ -20,7 +21,7 @@ from aiokafka.protocol.types import (
 )
 
 
-def test_create_message():
+def test_create_message() -> None:
     payload = b"test"
     key = b"key"
     msg = Message(value=payload, key=key, magic=0, attributes=0, crc=0)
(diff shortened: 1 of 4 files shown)


tests/test_protocol.py
```diff
@@ -283,9 +284,9 @@ def test_decode_fetch_response_partial():
             b"ar",  # Value (truncated)
         ]
     )
-    resp = FetchResponse[0].decode(io.BytesIO(encoded))
-    assert len(resp.topics) == 1
-    topic, partitions = resp.topics[0]
+    resp = FetchResponse_v0.decode(io.BytesIO(encoded))
+    assert len(resp.topics) == 1  # type: ignore[attr-defined]
+    topic, partitions = resp.topics[0]  # type: ignore[attr-defined]
```
ods, line 289: We should rather fix this problem in response class

tests/test_protocol.py
```diff
@@ -294,18 +295,18 @@ def test_decode_fetch_response_partial():
...
-        MetadataRequest[0](topicz="foo")
+        MetadataRequest_v0(topicz="foo")
 
 
-def test_struct_missing_kwargs():
-    fr = FetchRequest[0](max_wait_time=100)
-    assert fr.min_bytes is None
+def test_struct_missing_kwargs() -> None:
+    fr = FetchRequest_v0(max_wait_time=100)
+    assert fr.min_bytes is None  # type: ignore[attr-defined]
```
ods, line 306: Same here

tests/test_protocol_object_conversion.py
```diff
@@ -6,62 +8,119 @@
...
+                API_KEY = 0
+                API_VERSION = 0
+                RESPONSE_TYPE = Response  # To satisfy the Request ABC
+                SCHEMA = Schema(("myarray", Array(Int16)))
+        else:
+
+            class TestClass(Response):  # type: ignore[no-redef]
+                API_KEY = 0
+                API_VERSION = 0
+                SCHEMA = Schema(("myarray", Array(Int16)))
```
ods, line 88: Looks like the same code is repeated a lot of times. I think we can put it into generic function like this:
```python
def envelope_with_schema(superclass: ET, schema: Schema) -> ET:
    ...
```
Or even a parametrized factory fixture with single `schema` argument.

tests/test_protocol_object_conversion.py
```diff
@@ -1,67 +1,81 @@
...
+    else:
+
+        class TestClass(Response):  # type: ignore[no-redef]
             API_KEY = 0
             API_VERSION = 0
-            RESPONSE_TYPE = None  # To satisfy the Request ABC
-            SCHEMA = Schema(("myobject", Int16))
+            SCHEMA = schema
+
+    return TestClass  # type: ignore[return-value]
```
ods, line 27: Hmm, looks like mypy is right here, as the type we return is not the same. I don't know how to solve this in generic way, but we actually don't need to. Also, it's incorrect to use `issubclass` here, as we accept only `Request` and `Response`, but not subclasses.  Maybe this way?:

```python
def _make_test_class(base_cls: Type[Union[Request, Response]]) ->  Type[Union[Request, Response]]:
    if base_cls is Request:
        class TestRequest(Request):
            ...

        return TestRequest

    else:
        assert base_cls is Response

        class TestResponse(Response):
            ...

        return TestResponse
```

tests/test_protocol.py
dimastbk: I see these ways:
1. Use annotations on struct classes and generate __init__/attributes from these annotations (maybe use descriptor and dataclass/dataclass_transform)
2. Add annotations to class from schemas
3. Write mypy plugin like the existing mypy plugin for dataclasses

tests/test_protocol_object_conversion.py
dimastbk: I prefer generic because if we add annotations to the __init__ method, we should know the return value from class factory

dimastbk: Maybe use the version with generic (and # type: ignore[return-value]) and add a protocol like this? This solves the problem of differences between Response/Request.

```python
class ResponseRequestProtocol(Protocol):
    def __init__(self, *args: Any, **kwargs: Any) -> None: ...
    def to_object(self) -> Dict[str, Any]: ...
    def encode(self) -> bytes: ...
    def get_item(self, name: str) -> Any: ...
```

ods: > I prefer generic because if we add annotations to the init method, we should know the return value from class factory

I don't know how to express this with generic. It's not very good, but we can use `type: ignore` if we're sure our typing is right and there is a bug in mypy. But here I'm pretty sure that mypy is right and typing is incorrect: when `TestClass` is subclass of `Request` (or `Response`) this doesn't mean that `Type[TestClass]` is subclass of `Type[Request]` (or `Type[Response]`). And don't see how suggested protocol may solve this problem.

ods: BTW, do you have a Telegram account? Mine is [otkds](https://t.me/otkds), probably we might discuss ideas in realtime. (I'll remove this comment later to avoid spam)

PR #1006, opened 2024-05-06
add typing to aiokafka/coordinator/*

dimastbk:
### Changes

Fixes #

### Checklist


aiokafka/coordinator/assignors/sticky/sticky_assignor.py
```diff
@@ -70,43 +94,49 @@ class StickyAssignorUserDataV1(Struct):
...
         # a mapping of all consumers to all potential topic partitions that can be
         # assigned to them
-        self.consumer_to_all_potential_partitions = {}
+        self.consumer_to_all_potential_partitions: Dict[str, List[TopicPartition]] = {}
         # an ascending sorted set of consumers based on how many topic partitions are
         # already assigned to them
-        self.sorted_current_subscriptions = SortedSet()
+        self.sorted_current_subscriptions: SortedSet[
+            Tuple[str, Tuple[TopicPartition, ...]]
+        ] = SortedSet()
```
ods, line 122: _nit:_ Here and in other places, it would be nice to avoid breaking in inner expression when it's possible to break in outer, the current version is difficult to read: we have a statement `lvalue: type = expression` and the break is inside of "word" `type` instead of being between word. I saw this or similar problem was solved in some recent version of ruff. Maybe its update will be enough?

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
```diff
@@ -303,9 +338,10 @@ def _populate_sorted_partitions(self):
             # at this point, sorted_consumers contains an ascending-sorted list of
             # consumers based on how many valid partitions are currently assigned to
             # them
-            while sorted_consumers:
+            while item := sorted_consumers.last():
+                sorted_consumers.pop_last()
                 # take the consumer with the most partitions
```
ods, line 343: Now the comment is after the statement to which it applies

aiokafka/coordinator/assignors/abstract.py
```diff
@@ -37,8 +51,9 @@ def metadata(self, topics):
             MemberMetadata struct
         """
 
+    @classmethod
```
ods, line 54: (offtopic) Why is it class method? It's strange to have a shared state among all clients

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
```diff
@@ -170,14 +200,14 @@ def balance(self):
...
 
-    def get_final_assignment(self, member_id):
-        assignment = defaultdict(list)
+    def get_final_assignment(self, member_id: str) -> Dict[str, List[int]]:
+        assignment: Dict[str, List[int]] = defaultdict(list)
         for topic_partition in self.current_assignment[member_id]:
             assignment[topic_partition.topic].append(topic_partition.partition)
         assignment = {k: sorted(v) for k, v in assignment.items()}
-        return assignment.items()
+        return assignment
```
ods, line 208: This is incompatible change. Could you justify, why this way it's better?

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
dimastbk: Something like this was solved in ruff 3 or 4, but not for this case. May be use `# fmt: skip`?

aiokafka/coordinator/assignors/abstract.py
```diff
@@ -1,20 +1,33 @@
...
+    def name(self) -> str:
         """.name should be a string identifying the assignor"""
 
+    @classmethod
     @abc.abstractmethod
-    def assign(self, cluster, members):
+    def assign(
+        cls,
+        cluster: ClusterMetadata,
+        members: Dict[str, ConsumerProtocolMemberMetadata],
```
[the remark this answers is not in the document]
dimastbk: But it's like `return type` from https://typing.readthedocs.io/en/latest/source/best_practices.html#arguments-and-return-types (because `GroupCoordinator` always call this method with dict). But I watch only actual types, if this does not make sense, I will change this (and below) to protocols.

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
dimastbk: Reverted

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
ods: Not very important, at your discretion

aiokafka/coordinator/assignors/abstract.py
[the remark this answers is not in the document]
ods: I don't get it. Why is it like return type, when it _is_ argument type? Anyway, it's a question, I'm not sure what will be better. It's abstract class, so those variants are not fully compatible.

tests/coordinator/test_assignors.py
```diff
@@ -1,27 +1,36 @@
...
 
 
 def create_cluster(
-    mocker, topics, topics_partitions=None, topic_partitions_lambda=None
-):
+    mocker: MockerFixture,
+    topics: Set[str],
+    topics_partitions: Optional[Set[int]] = None,
+    topic_partitions_lambda: Optional[Callable[[str], Optional[Set[int]]]] = None,
+) -> MagicMock:
```
ods, line 33: Shouldn't we declare mocked type instead? Not sure whether this is correct though.

Makefile
```diff
@@ -17,6 +18,7 @@ FORMATTED_AREAS=\
 	tests/test_helpers.py \
 	tests/test_protocol.py \
 	tests/test_protocol_object_conversion.py \
+	tests/coordinator \
```
ods, line 21: _nit:_ Let's be consistent
```suggestion
	tests/coordinator/ \
```

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
```diff
@@ -548,7 +552,8 @@ def _perform_reassignments(
                     )
 
                 if (
-                    partition in self.previous_assignment
+                    consumer is not None
```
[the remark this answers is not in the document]
dimastbk: Because current_assignment is dict[str, list[TopicPartition]] and doesn't accept None key (`Invalid index type "Optional[str]" for "Dict[str, List[TopicPartition]]"; expected type "str"  [index]`). I don't see why the consumer can be None in the code. 


I didn't deep dive into the sticky assignment algorithm, but librdkafka uses assert, if `consumer` is None, see [librdkafka](https://github.com/confluentinc/librdkafka/blob/2587cac70f83fced42c51f921bed325a434f5bc7/src/rdkafka_sticky_assignor.c#L940C25-L940C34).

tests/coordinator/test_assignors.py
dimastbk: I think, because we return Mock, we should use mock as an annotation.

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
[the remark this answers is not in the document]
ods: Shouldn't we use `assert consumer is not None` instead of changing the branch condition?

tests/coordinator/test_assignors.py
ods: We may want to introspect the mock to check calls, so your annotation is correct

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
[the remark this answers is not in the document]
dimastbk: Done

aiokafka/coordinator/assignors/sticky/sticky_assignor.py
```diff
@@ -498,10 +557,9 @@ def _perform_reassignments(self, reassignable_partitions):
                     )
                 # the partition must have a current consumer
                 consumer = self.current_partition_consumer.get(partition)
-                if consumer is None:
-                    log.error(
-                        "Expected partition %r to be assigned to a consumer", partition
-                    )
+                assert (
+                    consumer is not None
+                ), f"Expected partition {partition!r} to be assigned to a consumer"
```
ods, line 562: Sorry for changing my mind. Looks like my suggestion may potentially kill coordination task. Let's return to your variant, but reword it this way:
```suggestion
                if consumer is None:
                    log.error(
                        "Expected partition %r to be assigned to a consumer", partition
                    )
                    continue
```
Please recheck that its behaviour is equivalent to what we had before.

PR #1007, opened 2024-05-09
Fix unintended cancellation of fetcher task

### LINKED ISSUE #983: Consumer stopped consuming, task Fetcher._fetch_task has finished
**Describe the bug**
Under unclear conditions, our consumer has stopped receiving messages.
We investigated, when it happens, `_fetch_task` is finished, despite the `Fetcher` object is still alive (`Fetcher.close()` was not called).

```
 '_fetch_task': <Task finished name='Task-29' coro=<Fetcher._fetch_requests_routine() done, defined at /opt/poetry/venv/lib/python3.12/site-packages/aiokafka/consumer/fetcher.py:457> result=None>,
``` 
We suspect the issue might be in the `_fetch_requests_routine()`, that it incorrectly handles `CancelledError`.
According to the python documentation, `CancelledError` shall be reraised, while the above routine just suppress it.

 - aiokafka version: 0.10.0
 - Kafka Broker version: Confluent Kafka
 - Other information (Confluent Cloud version, etc.):

apmorton:
### Changes

Fixes #983

It's not entirely clear to me why the change from `asyncio.wait_for` to `util.wait_for` caused a behavior change here, but this code was always broken in the case where `task.cancel()` was called.

### Checklist

### DIFF
--- a/aiokafka/consumer/fetcher.py
+++ b/aiokafka/consumer/fetcher.py
@@ -512,10 +512,8 @@ class Fetcher:
                         # cancellation
                         if not task.done():
                             task.cancel()
-                        try:
+                        with contextlib.suppress(asyncio.CancelledError):
                             await task
-                        except asyncio.CancelledError:
-                            pass
                     self._pending_tasks.clear()
                     self._records.clear()
 
--- a/aiokafka/consumer/fetcher.py
+++ b/aiokafka/consumer/fetcher.py
@@ -512,7 +512,10 @@ class Fetcher:
                         # cancellation
                         if not task.done():
                             task.cancel()
-                        await task
+                        try:
+                            await task
+                        except asyncio.CancelledError:
+                            pass
                     self._pending_tasks.clear()
                     self._records.clear()
 


aiokafka/consumer/fetcher.py
```diff
@@ -512,7 +512,10 @@ def on_done(fut, self=self):
                         # cancellation
                         if not task.done():
                             task.cancel()
-                        await task
+                        try:
+                            await task
```
ods, line 516: Let's imagine, that `_fetch_requests_routine` is cancelled while awating the task here. Then instead of terminating it will continue infinite loop causing routing to hang. Right?

ods: I did some research and it looks like there is no a single case in Python code base where they unconditionally swallow `CancelledError` exception. But it looks like a hard-to-hit race and we already suppress `CancelledError` in other places in `aiokafka`, so let's go this way for now and resolve it consistently later.

aiokafka/consumer/fetcher.py
```diff
@@ -512,7 +512,10 @@ def on_done(fut, self=self):
...
+                        except asyncio.CancelledError:
+                            pass
```
ods, line 518:
```suggestion
                        with contextlib.suppress(asyncio.CancelledError):
                            await task
```

Issue #830, opened 2022-04-14
Error logs are continuously printed when disconnected from Kafka


tufbel:
```python
      consumer_ = aiokafka.AIOKafkaConsumer(
          *topics,
          bootstrap_servers=f"192.168.1.100:9092",
      )
      await consumer_.start()
      try:
          async for msg in consumer_:
              print(msg)
      finally:
          await consumer_.stop()
```
`aiokafka` outputs a lot of error logs when I disconnect from 192.168.1.100 :

```
ERROR    | aiokafka.consumer.fetcher:_proc_fetch_request:658 - Failed fetch messages from 3: KafkaConnectionError: Connection at 10.64.73.35:9094 closed
ERROR    | aiokafka.client:_get_conn:460 - Unable connect to node with id 3: [WinError 1232]
```

Issue #496, opened 2019-04-11
Too many reqs on kafka not available 


lud4ik:
I have code with simple consumer. I undestand that most of time at least 1 of nodes will be available. But on kafka stop it send too many requests and uses up to 100% cpu
```
2019-04-11 12:06:55,732 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,732 Unable to update metadata from [0]
2019-04-11 12:06:55,732 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,732 Failed fetch messages from 0: NodeNotReadyError: Attempt to send a request to node which is not ready (node id 0).
1
2019-04-11 12:06:55,733 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,733 Unable to update metadata from [0]
2019-04-11 12:06:55,734 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,734 Failed fetch messages from 0: NodeNotReadyError: Attempt to send a request to node which is not ready (node id 0).
1
2019-04-11 12:06:55,734 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,734 Unable to update metadata from [0]
2019-04-11 12:06:55,735 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,735 Failed fetch messages from 0: NodeNotReadyError: Attempt to send a request to node which is not ready (node id 0).
1
2019-04-11 12:06:55,736 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,736 Unable to update metadata from [0]
2019-04-11 12:06:55,736 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,736 Failed fetch messages from 0: NodeNotReadyError: Attempt to send a request to node which is not ready (node id 0).
1
2019-04-11 12:06:55,737 Unable connect to node with id 0: [Errno 111] Connect call failed ('10.0.3.2', 9092)
2019-04-11 12:06:55,737 Unable to update metadata from [0]
[...]

PR #1014, opened 2024-06-12
Added delay to retry connection

camcima:
When a consumer connection breaks, aiokafka goes into a loop trying to reconnect every 1ms, generating a mountain of logs. This PR introduces a delay between connection attempts with the configured `retry_backoff_ms` parameter.

Addresses issues #496 and #830.


jainal09:
Is it possible to have the functionality at the user level where we can catch these network connection errors. Because even with exponential back off it will keep retrying indefinitely

mashkovd:
[#496](https://github.com/aio-libs/aiokafka/discussions/949)
I wrote about this issues.
@ods can you review it? thx

PR #1024, opened 2024-07-02
 add typing to aiokafka/cluster.py

dimastbk:
### Changes

Fixes #

### Checklist


aiokafka/cluster.py
```diff
@@ -38,6 +38,9 @@
 )
 from aiokafka.structs import BrokerMetadata, PartitionMetadata, TopicPartition
 
+if TYPE_CHECKING:
+    from aiokafka.client import CoordinationType
```
ods, line 42: What do you think about moving `ConnectionGroup` and `CoordinationType` to `structs.py` instead? Also, I see that they just namespace classes, so it's meaningless to use them in type declarations. Probably we need to make them enums?

jackgene: Alternatively, create `aiokafka/enums.py` for common enums.

Issue #1050, opened 2024-10-07
[MINOR] structs.BrokerMetadata#nodeId incorrectly typed


jackgene:
**Describe the bug**
[aiokafka.structs.BrokerMetadata#nodeId](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/structs.py#L29) is typed as an `int`, however it is sometimes assigned `str` values:

- It is assigned a `str` [here](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/cluster.py#L77-L78)
- Also assigned a `str` [here](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/cluster.py#L350-L351)
- It is assigned an `int` [here](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/cluster.py#L390) (the value is passed in to `add_coordinator(...)` [here](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/client.py#L696), the value of `resp.coordinator_id` is an integer, as defined [here](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/protocol/coordination.py#L10), and [here](https://github.com/aio-libs/aiokafka/blob/01c60cd9c45e975aee8d4932247108a9e639d026/aiokafka/protocol/coordination.py#L23))

**Expected behaviour**
`aiokafka.structs.BrokerMetadata#nodeId` is correctly typed as a `typing.Union[int, str]`

Value reads do not seem to care too much about the type of this field, as long as:
- It is unique
- It can be used as a `dict` key

So no further changes should be needed

Alternative is to make it a `str` field, and convert `int` to `str` where `int` values are currently being assigned.

**Environment (please complete the following information):**
N/A this is a typing bug, I generally use Pyright to type check.

**Additional observations**
Also note that the name `nodeId` is camel-case, whereas Python prefers (and the rest of the code-base are) snake-case. I will provide a PR to resolve the above issue shortly. I am happy to rename `nodeId` to `node_id` as well, if that is desired, but I do not plan to, to minimize the changes introduced by the PR.

PR #1051, opened 2024-10-07
Updated BrokerMetadata#nodeId type from `int` to `int | str`.

jackgene:
### Changes

Fixes https://github.com/aio-libs/aiokafka/issues/1050

Preferred (less invasive) fix to issue above, changes the type of `aiokafka.structs.BrokerMetadata#nodeId` from `int` to `int | str`. No functionality change.

Use this or https://github.com/aio-libs/aiokafka/pull/1052, but not both.

### Checklist

The check list mentions a `CHANGES` folder, but I do not see one, am I supposed to create one for each PR?

### DIFF
--- a/aiokafka/structs.py
+++ b/aiokafka/structs.py
@@ -1,6 +1,7 @@
+from __future__ import annotations
 from collections.abc import Sequence
 from dataclasses import dataclass
-from typing import Generic, List, NamedTuple, Optional, Sequence, Tuple, TypeVar, Union
+from typing import Generic, NamedTuple, Optional, Sequence, TypeVar
 
 from aiokafka.errors import KafkaError
 
@@ -27,7 +28,7 @@ class TopicPartition(NamedTuple):
 class BrokerMetadata(NamedTuple):
     """A Kafka broker metadata used by admin tools"""
 
-    nodeId: Union[int, str]
+    nodeId: int | str  # FIXME consider updating implementation
     "The Kafka broker id"
 
     host: str
--- a/aiokafka/structs.py
+++ b/aiokafka/structs.py
@@ -1,6 +1,6 @@
 from collections.abc import Sequence
 from dataclasses import dataclass
-from typing import Generic, NamedTuple, Optional, TypeVar
+from typing import Generic, List, NamedTuple, Optional, Sequence, Tuple, TypeVar, Union
 
 from aiokafka.errors import KafkaError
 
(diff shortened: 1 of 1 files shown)


ods:
Hi @jackgene, thank you for the contribution. You're right, the the type annotation doesn't match the code. But I'm not sure we need to fix the type annotation here and not the code.

jackgene:
> Hi @jackgene, thank you for the contribution. You're right, the the type annotation doesn't match the code. But I'm not sure we need to fix the type annotation here and not the code.

Yup, and I did offer two PRs, though they both touch the type annotation.

As I understand it, in the aiokafka model, `node_id` is used for two things:
- `int` ID, as assigned by Kafka
- Synthetically generated `str` ID to denote `broker`/`coordinator` by the aiokafka library

We obviously do not want the two to collide, and it won't in the current implementation, as one is an `int` and the other is a `str`. We can keep things the way they are, and make it explicit that the type of `node_id` is `int | str`.

Alternatively, we can make it a single type, `str` (in my other PR), makes it easy to distinguish between the two classes of ids (numeric strings, vs strings prefixed by `broker-`/`coordinator-`).

We could also make it an `int`, but this require knowledge of how Kafka assigns IDs that I'm unfortunately not familiar with. If Kafka only assigns positive `int` for instance, we could do negative `int`s to represent `broker`/`coordinator`s.

So IMO, changing the type annotation is the less invasive option.

ods:
Ok, as a temporary solution it's reasonable to keep it as-is. Let's go this way. Please rebase it and add a FIXME comment explaining why it's defined this way

aiokafka/structs.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import annotations
```
ods, line 1: There is a couple of linting errors
```suggestion
from __future__ import annotations

```

aiokafka/structs.py
```diff
@@ -27,7 +28,7 @@ class TopicPartition(NamedTuple):
 class BrokerMetadata(NamedTuple):
     """A Kafka broker metadata used by admin tools"""
 
-    nodeId: int
+    nodeId: int | str  # FIXME consider updating implementation (https://github.com/aio-libs/aiokafka/issues/1050)
```
ods, line 31:
```suggestion
    # FIXME: consider updating implementation (https://github.com/aio-libs/aiokafka/issues/1050)
    nodeId: int | str
```

ods (changes requested):

PR #1068, opened 2024-10-30
Fix incomplete documentation for AIOKafkaConsumer.offset_for_times

### LINKED ISSUE #1067: Consumer.offsets_for_times return type documentation is incomplete.
**Describe the bug**
Currently the return type documentation for the `offsets_for_times` function is:
```
Returns:
            dict(TopicPartition, OffsetAndTimestamp): mapping from
            partition to the timestamp and offset of the first message with
            timestamp greater than or equal to the target timestamp.
```

See https://aiokafka.readthedocs.io/en/stable/api.html#aiokafka.AIOKafkaConsumer.offsets_for_times
The most recent kafka javadocs for this function state:

```
a mapping from partition to the timestamp and offset of the first message with timestamp greater than
or equal to the target timestamp. null will be returned for the partition if there is no such message.
```
See https://kafka.apache.org/38/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#offsetsForTimes(java.util.Map)

It is missing the returns None if there is no message that fulfils the timestamp condition. This is extra confusing because the function description includes a section stating:
```
If the message format version in a partition is before 0.10.0, i.e.
the messages do not have timestamps, ``None`` will be returned for that
partition.
```

The docs now say that `None` only happens if the cluster is dealing with an outdated message format. Rather than it being part of the normal interface.

**Expected behaviour**
The `offsets_for_times` documentations is brought in line with the kafka javadocs and the actual behaviour of the function.

 - aiokafka version: 0.12.0
 - Kafka Broker version (`kafka-topics.sh --version`): Not relevant
 - Other information (Confluent Cloud version, etc.): Not relevant

**Reproducible example**
Go to https://github.com/aio-libs/aiokafka/blob/master/aiokafka/consumer/consumer.py#L906 and see the docstring.

jzvandenoever:
### Changes

Fixes #1067 

Updates the return description to be in line with the behaviour of the
offset_for_times function and the official Kafka documentation.

### DIFF
--- a/aiokafka/consumer/consumer.py
+++ b/aiokafka/consumer/consumer.py
@@ -927,7 +927,8 @@ class AIOKafkaConsumer:
         Returns:
             dict(TopicPartition, OffsetAndTimestamp): mapping from
             partition to the timestamp and offset of the first message with
-            timestamp greater than or equal to the target timestamp.
+            timestamp greater than or equal to the target timestamp. None will
+            be returned for the partition if there is no such message.
 
         Raises:
             ValueError: If the target timestamp is negative


ods:
Hi @jzvandenoever, thank you for the contribution! You're right, and we even have [a test for such case](https://github.com/ods/aiokafka/blob/master/tests/test_consumer.py#L1498).

PR #1069, opened 2024-10-31
Update Java Client API reference

emmanuel-ferdman:
### Changes

Small PR - fixes the Java Client API reference

### Checklist

### DIFF
--- a/docs/kafka-python_difference.rst
+++ b/docs/kafka-python_difference.rst
@@ -103,7 +103,7 @@ this value to ``max_poll_interval_ms``, same as in Java client.
 Prefetching is more sophisticated
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


ods:
Hi @emmanuel-ferdman,
What problem do you see in the current version? This reference in [the generated doc](https://aiokafka.readthedocs.io/en/stable/kafka-python_difference.html#why-do-we-need-another-library) works fine for me.

emmanuel-ferdman:
@ods, it leads to error page at `https://aiokafka.readthedocs.io/en/stable/JavaClientAPI`:
image

On that page, there are two links that should lead to `https://kafka.apache.org/documentation/#api`. The first one works, while the second one (which unlike the first one, uses a custom clickable text with a cross-reference label) does not work.

ods:
Oh, I see. Thank you for the fix!

PR #1074, opened 2024-11-15
Added type annotation to public classes.

jackgene:
### Changes
Adds type annotation to public classes, and `py.typed` so that client code can type check against this library.

Fixes https://github.com/aio-libs/aiokafka/issues/980 (kind of)

Note that the issue in question seems to be more ambitious in that it aims to provide type information to the entire library:
1. Allowing bugs to be discovered
2. Allowing developers consuming this library to have better type information to work with. 

This PR only aims to solve the latter.

## Details about this PR:
- Using Python 3.10+ typing, which is available in 3.9 via `from __future__ import annotations`
	- `x | None` instead of `Optional[x]`
	- `x | y` instead of `Union[x, y]`
	- `list[x]`/`set[x]`/`dict[x, y]` instead of `typing.List[x]`/`typing.Set[x]`/`typing.Dict[x, y]`
- For the most part, all I'm doing is explicitly annotate methods as documented in docstring
- There are special cases though where the docstring had to be updated, e.g.,:
	- docstring says `list`, but varargs/default value is `tuple`
	- docstring doesn't say optional, but default value is `None`

Note that this is introducing a breaking change that may affect a small minority of users. The key/value serializer/deserializer functions no longer optional (with the identity function defined as the default parameter). This is so that the generic parameters `KT`/`VT` can be inferred.

This means that:
- The (de)serialization behavior is identical at runtime
- As long as clients do not explicitly set `None` as their key/value serializer/deserializer function, there should be no change.

However, if a client is explicitly passing in `None`, e.g., with `AIOKafkaConsumer`:
```python
# OK, most people probably do this anyway, this is from the README
consumer = AIOKafkaConsumer(
    'my_topic', 'my_other_topic',
    bootstrap_servers='localhost:9092',
    group_id='my-group'
)

# Not OK, key_deserializer/value_deserializer cannot be None
consumer = AIOKafkaConsumer(
    'my_topic', 'my_other_topic',
    bootstrap_servers='localhost:9092',
    group_id='my-group',
    key_deserializer=None,
    value_deserializer=None
)
```

Or with `AIOKafkaProducer`:
```python
# OK, most people probably do this anyway, this is from the README
AIOKafkaProducer(
    bootstrap_servers='localhost:9092'
)

# Not OK, key_deserializer/value_deserializer cannot be None
AIOKafkaProducer(
    bootstrap_servers='localhost:9092'
    key_serializer=None,
    value_serializer=None
)
```

The type checker will reject it.
image

You can ignore the type checker though, and the application should work the same as before, as this PR does not change any runtime logic.

### Checklist

### DIFF
--- a/aiokafka/abc.py
+++ b/aiokafka/abc.py
@@ -1,5 +1,7 @@
 import abc
 
+from aiokafka.structs import TopicPartition
+
 
 class ConsumerRebalanceListener(abc.ABC):
     """
@@ -45,7 +47,7 @@ class ConsumerRebalanceListener(abc.ABC):
     """
 
     @abc.abstractmethod
-    def on_partitions_revoked(self, revoked):
+    def on_partitions_revoked(self, revoked: list[TopicPartition]) -> None:
         """
         A coroutine or function the user can implement to provide cleanup or
         custom state save on the start of a rebalance operation.
@@ -65,7 +67,7 @@ class ConsumerRebalanceListener(abc.ABC):
         """
 
     @abc.abstractmethod
-    def on_partitions_assigned(self, assigned):
+    def on_partitions_assigned(self, assigned: list[TopicPartition]) -> None:
         """
         A coroutine or function the user can implement to provide load of
         custom consumer state or cache warmup on completion of a successful
@@ -103,7 +105,7 @@ class AbstractTokenProvider(abc.ABC):
     """
 
     @abc.abstractmethod
-    async def token(self):
+    async def token(self) -> None:
         """
         An async callback returning a :class:`str` ID/Access Token to be sent to
         the Kafka client. In case where a synchronous callback is needed,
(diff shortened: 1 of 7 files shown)


aiokafka/abc.py
```diff
@@ -103,7 +105,7 @@ class AbstractTokenProvider(abc.ABC):
     """
 
     @abc.abstractmethod
-    async def token(self):
+    async def token(self) -> None:
```
shuckc, line 108: Shouldn't this be `-> str`?

jackgene: Good catch. Based on the doc comment, I think you are right. I believe these are auto-generated though, so let me look to see if there are any concrete implementations, and what those are returning.

jackgene: Fixed.

PR #1076, opened 2024-11-18
Use SPDX license identifier

cdce8p:
### Changes
Use the SPDX license identifier as recommended by [PEP 639](https://peps.python.org/pep-0639/).
https://spdx.org/licenses/Apache-2.0.html

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,17 +1,16 @@
 [build-system]
-requires = ["setuptools >=61", "wheel", "Cython >=3.0.5"]
+requires = ["setuptools >=77", "Cython >=3.0.5"]
 
 [project]
 name = "aiokafka"
 description = "Kafka integration with asyncio"
 readme = "README.rst"
 requires-python = ">=3.9"
-license = { file = "LICENSE" }
+license = "Apache-2.0"
 authors = [
     { name = "Andrew Svetlov", email = "andrew.svetlov@gmail.com" },
 ]
 classifiers = [
-    "License :: OSI Approved :: Apache Software License",
     "Intended Audience :: Developers",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.9",
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -109,7 +109,7 @@ jobs:
 
       - name: Install python dependencies
         run: |
-          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install --upgrade pip setuptools
           python -m pip install -r requirements-win-test.txt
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -6,7 +6,7 @@ name = "aiokafka"
 description = "Kafka integration with asyncio"
 readme = "README.rst"
 requires-python = ">=3.9"
-license = { file = "LICENSE" }
+license = { text = "Apache-2.0" }
```
ods, line 9: According to PEP, this ways is [deprecated](https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys):

> Table values for the `license` key in the `[project]` table, including the `text` and `file` table subkeys, are now deprecated.

```suggestion
license = "Apache-2.0"
```

cdce8p: Yeah. Unfortunately setuptools doesn't support the new syntax yet. Using `license.text` until then will make updating later a bit easier and also help with readability on PyPI.

**before**

Screenshot 2024-11-19 at 08 37 26

**after**

Screenshot 2024-11-19 at 08 38 02

ods: Probably it's time to switch to `hatch` (need to figure out how to deal with `cython` to do this). If not, I'd rather wait until `setuptools` is fixed.

cdce8p: Setuptools `v77` was released this week which adds full support for PEP 639. Updated the PR to use the new syntax. It should be ready now.

pyproject.toml
```diff
@@ -1,17 +1,16 @@
 [build-system]
-requires = ["setuptools >=61", "wheel", "Cython >=3.0.5"]
+requires = ["setuptools >=77", "Cython >=3.0.5"]
```
ods, line 2: Are you sure `wheel` package is not needed? It is not listed in basic dependencies of `setuptools`, only in optional (extra)

cdce8p: > Are you sure `wheel` package is not needed?

With `v70.1` setuptools started vendoring `wheel` and modifying it for its own use. Since then it has not relied on the `wheel` package itself.
https://setuptools.pypa.io/en/latest/history.html#v70-1-0

The official docs also only mention `setuptools` for `build-system.requires`.
https://setuptools.pypa.io/en/latest/build_meta.html#how-to-use-it

Furthermore, I doubled checked that `python -m build` still works.

ods: If we remove it from here, we have to remove it from requirements*.txt files too.

ods: Sorry, you're right, it was installed directly in CI.

PR #1105, opened 2025-05-06
KIP-368: Allow SASL Connections to Periodically Re-Authenticate

### LINKED ISSUE #1015: [QUESTION] Is there support to refresh OAuth tokens automatically based on token expiration duration?
Hi,

We're using OAUTHBEARER sasl_mechanism to connect to AWS MSK with MSKAuthTokenProvider generated tokens. The tokens have an expiration duration. Is there any way to make the Kafka clients refresh the connections before the token expires? Right now, we observe that a connection is established. Later, there will be a Kafka connection failure because the connection is closed by the broker.

### LINKED ISSUE #1080: Consumer/channel does not reauthenticate based on SaslAuthenticateResponse
**Describe the bug**
When a consumer's connection outlives its SASL authentication (in this case an OAuth token), then consumer will fail to fetch new records because the broker no longer accepts any requests other than `SaslHandshakeRequest` and `SaslAuthenticateRequest`. 

As an example, if tokens have a valid duration of 5 minutes, the consumer will run fine for 5 minutes after which it will begin failing when it sends new requests. For example, this `HeartbeatRequest` for the group coordinator fails because the channel's session is expired:
```
ERROR:aiokafka.consumer.group_coordinator:Error sending HeartbeatRequest_v1 to node 2 [KafkaConnectionError: Connection at my.redacted-kafka.broker closed] -- marking coordinator dead
```

I suspect this is also an issue for admin and producer clients.

**Expected behaviour**
I expect the consumer to reauthenticate before its session expires.

 - aiokafka version (`python -c "import aiokafka; print(aiokafka.__version__)"`): 0.12.0
 - Kafka Broker version (`kafka-topics.sh --version`): 3.7.0
 - Other information (Confluent Cloud version, etc.): Self-hosted Kafka with Strimzi OAuth authentication plugin

**Reproducible example**

The problem can be reproduced with a simple consumer such as this together with a Kafka broker configured to use secured OAuth SASL authentication for example using Strimzi. If it could be of value, I will try to create a Dockerfile to setup such a broker.

```python
import asyncio
from aiokafka.abc import AbstractTokenProvider
from azure.identity import DefaultAzureCredential
import ssl
def create_ssl_context():
    """
    Creates and configures an SSL context for secure connections over TLS.
    Used in communication with the Kafka broker in aiokafka.Consumer.
    Returns:
        ssl.SSLContext: Configured SSL context for use with Kafka or other secure connections.
    """
    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
    ssl_context.load_default_certs()
    return ssl_context
class KafkaOAuthBearerTokenProvider(AbstractTokenProvider):
    async def token(self):
        # logic for fetching token
        ...
async def consume():
    # Kafka consumer configuration with SASL authentication
    consumer = AIOKafkaConsumer(
        'sometopic',
        bootstrap_servers='my.redacted-kafka.broker,my.redacted-kafka.broker2,my.redacted-kafka.broker3',
        auto_offset_reset='earliest',
        security_protocol='SASL_PLAINTEXT', 
        sasl_mechanism='OAUTHBEARER',
        ssl_context=create_ssl_context(),
        sasl_oauth_token_provider= KafkaOAuthBearerTokenProvider(),
        group_id= "someconsumergroup"
    )
    # Start the consumer
    await consumer.start()
    try:
        # Consume messages
        async for msg in consumer:
            print(f"Consumed message: {msg.value.decode('utf-8')} from partition: {msg.partition} and offset: {msg.offset} and header: {msg.headers}")
    finally:
        # Ensure the consumer is stopped gracefully
        await consumer.stop()
```

kprzybyla:
### Changes

Fixes #1080
Fixes #1015

I recently discovered that `aiokafka` does not implement `KIP-368`, which is essential for the `OAUTHBEARER` authorization, for example, when using `AWS MSK`. This was not so obvious to find out for me as someone who does not know Kafka internals and just wants to use a certain authorization method, because according to the `KIP-368`, the server will just break the connection to any client for which the token has expired. So, from the client side, this looks like the Kafka server is terminating the connection for no reason. This was especially mindboggling because the [Amazon MSK library](https://github.com/aws/aws-msk-iam-sasl-signer-python) "hardcodes" the [session timeout to be 15 minutes](https://github.com/aws/aws-msk-iam-sasl-signer-python/blob/a1bbf3c19eaff152051188bf2344e14727eacb2c/aws_msk_iam_sasl_signer/MSKAuthTokenProvider.py#L18), while the token is actually granted with a 1-hour expiration time, and this did not help to make the 1-hour error interval noticeable.

Also, Amazon MSK library describes in their ["Get Started"](https://github.com/aws/aws-msk-iam-sasl-signer-python?tab=readme-ov-file#get-started) how to use it with the `kafka-python` library, but this library does not implement `KIP-368` either (see https://github.com/dpkp/kafka-python/issues/2205), which leads me to believe that not that many people is aware how problematic this is when you are trying to built something reliable. This also happens to be the case for the `confluent-kafka-python` (see https://github.com/confluentinc/confluent-kafka-python/issues/1485), which is also used in the Amazon MSK example.

Anyway, after I figured out what was happening, I created a quick and dirty patch for `aiokafka` with a workaround that closed connections that were about to expire, and this got rid of all connection drops from the Kafka server that I was experiencing previously. This "patched" version is now used in our production in the company I worked for, and it works flawlessly, but due to certain implementation shortcuts, it generates unharmful but still annoying errors. So, after it was proven that this solved the issue we were battling for a long time, I decided to contribute a polished solution so that nobody else has to go through the same frustration I went through to get this working reliably.

The `OAUTHBEARER` SASL mechanism was also painful to configure in the Kafka server since it is implemented differently than other SASL mechanisms. Setting up `OAUTHBEARER` with other SASL mechanisms is impossible, at least I did not find a way to make it work. Because of this, a separate Docker container runs specifically for `OAUTHBEARER` tests.

Also, for anyone reading this who will use AWS MSK, please remember to set a static value for `AWS_ROLE_SESSION_NAME` (different for each client), otherwise, the re-authentication will fail since by default, the session name will contain the current timestamp at the end, and the session name must stay the same for a given client.

Java reference implementation of the same feature: https://github.com/apache/kafka/pull/5582

### Checklist

- [X] I think the code is well written
- [X] Unit tests for the changes exist
- [X] Documentation reflects the changes

### DIFF
--- a/aiokafka/conn.py
+++ b/aiokafka/conn.py
@@ -44,6 +44,9 @@ DEFAULT_KAFKA_PORT = 9092
 
 READER_LIMIT = 2**16
 SASL_QOP_AUTH = 1
+SASL_REQUEST_API_KEYS = frozenset(
+    request.API_KEY for request in (*SaslHandShakeRequest, *SaslAuthenticateRequest)
+)
 
 
 class CloseReason(IntEnum):
@@ -116,6 +119,24 @@ async def create_conn(
     return conn
 
 
+def calculate_sasl_reauthentication_time(session_lifetime_ms: int) -> int:
+    """
+    Calculates the SASL session re-authentication time following
+    the Java Kafka implementation from SaslClientAuthenticator.java
+    ReauthInfo#setAuthenticationEndAndSessionReauthenticationTimes.
+
+    The re-authentication factor is calculated by choosing random value
+    between 0.85 and 0.95, which accounts for both network latency and clock
+    drift as well as potential jitter which may cause re-authentication storm
+    across many channels simultaneously.
+    """
+
+    reauthentication_time_factor: float = random.uniform(0.85, 0.95)
+    expiration_time: float = (session_lifetime_ms * reauthentication_time_factor) / 1000
+
+    return int(expiration_time)
+
+
 class AIOKafkaProtocol(asyncio.StreamReaderProtocol):
     def __init__(self, closed_fut, *args, loop, **kw):
         self._closed_fut = closed_fut
@@ -184,6 +205,9 @@ class AIOKafkaConnection:
         self._sasl_kerberos_service_name = sasl_kerberos_service_name
         self._sasl_kerberos_domain_name = sasl_kerberos_domain_name
         self._sasl_oauth_token_provider = sasl_oauth_token_provider
+        self._sasl_reauthentication_task = None
+        self._sasl_reauthentication_done = asyncio.Event()
+        self._sasl_reauthentication_done.set()
 
         # Version hint is the version determined by initial client bootstrap
         self._version_hint = version_hint
@@ -358,6 +382,16 @@ class AIOKafkaConnection:
                     raise exc
                 auth_bytes = resp.sasl_auth_bytes
 
+                if (
+                    hasattr(resp, "session_lifetime_ms")
+                    and resp.session_lifetime_ms != 0
+                ):
+                    self._sasl_reauthentication_task = (
+                        self._create_sasl_reauthentication_task(
+                            resp.session_lifetime_ms
+                        )
+                    )
+
         if self._sasl_mechanism == "GSSAPI":
             log.info("Authenticated as %s via GSSAPI", self.sasl_principal)
         elif self._sasl_mechanism == "OAUTHBEARER":
@@ -368,6 +402,69 @@ class AIOKafkaConnection:
                 self._sasl_plain_username,
                 self._sasl_mechanism,
             )
+
+    def _create_sasl_reauthentication_task(
+        self, session_lifetime_ms: int
+    ) -> asyncio.Task:
+        self_ref = weakref.ref(self)
+        timeout = calculate_sasl_reauthentication_time(session_lifetime_ms)
+
+        log.info(
+            "SASL re-authentication required after %ds for connection %s:%s",
+            timeout,
+            self._host,
+            self._port,
+        )
+
+        sasl_reauthentication_task = create_task(
+            self._sasl_reauthentication(self_ref, timeout)
+        )
+        sasl_reauthentication_task.add_done_callback(
+            functools.partial(self._on_sasl_reauthentication_task_error, self_ref)
+        )
+
+        return sasl_reauthentication_task
+
+    @staticmethod
+    async def _sasl_reauthentication(
+        self_ref: weakref.ReferenceType["AIOKafkaConnection"],
+        sasl_reauthentication_time: int,
+    ) -> None:
+        self = self_ref()
+
+        if self is None:
+            return
+
+        await asyncio.sleep(sasl_reauthentication_time)
+        self._sasl_reauthentication_done.clear()
+
+        await self._do_sasl_handshake()
+        self._sasl_reauthentication_done.set()
+
+        log.info(
+            "SASL re-authentication complete for connection %s:%s",
+            self._host,
+            self._port,
+        )
+
+    @staticmethod
+    def _on_sasl_reauthentication_task_error(
+        self_ref: weakref.ReferenceType["AIOKafkaConnection"],
+        sasl_reauthentication_task: asyncio.Task,
+    ) -> None:
+        if sasl_reauthentication_task.cancelled():
+            return
+
+        try:
+            sasl_reauthentication_task.result()
+        except BaseException as exc:
+            if not isinstance(exc, (OSError, EOFError, ConnectionError)):
+                log.exception("Unexpected exception in AIOKafkaConnection")
+
+            self = self_ref()
+
+            if self is not None:
+                self.close(reason=CloseReason.AUTH_FAILURE, exc=exc)
 
     def authenticator_plain(self):
         return SaslPlainAuthenticator(
@@ -458,6 +555,18 @@ class AIOKafkaConnection:
                 f"No connection to broker at {self._host}:{self._port}"
             )
 
+        if (
+            self._sasl_reauthentication_done.is_set()
+            or request.API_KEY in SASL_REQUEST_API_KEYS
+        ):
+            return self._send(request=request, expect_response=expect_response)
+
+        return self._send_after_sasl_reauthentication(
+            request=request,
+            expect_response=expect_response,
+        )
+
+    def _send(self, request, expect_response=True):
         correlation_id = self._next_correlation_id()
         header = request.build_request_header(
             correlation_id=correlation_id, client_id=self._client_id
@@ -481,6 +590,11 @@ class AIOKafkaConnection:
             (correlation_id, request, fut),
         )
         return wait_for(fut, self._request_timeout)
+
+    async def _send_after_sasl_reauthentication(self, request, expect_response):
+        await self._sasl_reauthentication_done.wait()
+
+        return await self._send(request=request, expect_response=expect_response)
 
     def _send_sasl_token(self, payload, expect_response=True):
         if self._writer is None:
@@ -529,6 +643,13 @@ class AIOKafkaConnection:
                 self._on_close_cb = None
         if self._idle_handle is not None:
             self._idle_handle.cancel()
+
+        if (
+            self._sasl_reauthentication_task is not None
+            and not self._sasl_reauthentication_task.done()
+        ):
+            self._sasl_reauthentication_task.cancel()
+            self._sasl_reauthentication_task = None
 
         # transport.close() will close socket, but not right ahead. Return
         # a future in case we need to wait on it.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,13 @@ Changelog
 0.13.0 (????-??-??)
 ===================
 
+New features:
+
+* Allow SASL Connections to Periodically Re-Authenticate (`KIP-368`_) (pr #1105 by @kprzybyla)
+
+.. _KIP-368: https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate
+
+
 Improved Documentation:
 
 * Fix incomplete documentation for `AIOKafkaConsumer.offset_for_times``
--- a/docker/scripts/start-kafka.sh
+++ b/docker/scripts/start-kafka.sh
@@ -1,12 +1,12 @@
-#!/bin/sh
+#!/bin/bash
 
-OPTIONS=""
+OPTIONS=()
 PATH="$HOME/bin:$PATH"
 
 # Configure the default number of log partitions per topic
 if [ ! -z "$NUM_PARTITIONS" ]; then
     echo "default number of partition: $NUM_PARTITIONS"
-    OPTIONS="$OPTIONS --override num.partitions=$NUM_PARTITIONS"
+    OPTIONS+=("--override" "num.partitions=$NUM_PARTITIONS")
 fi
 
 # Set the external host and port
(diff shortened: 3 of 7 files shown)


aiokafka/conn.py
```diff
@@ -369,6 +403,69 @@ async def _do_sasl_handshake(self):
...
+        )
+
+        return sasl_reauthentication_task
+
+    @staticmethod
+    async def _sasl_reauthentication(
+        self_ref: weakref.ReferenceType["AIOKafkaConnection"],
+        sasl_reauthentication_time: int,
+    ) -> None:
+        self = self_ref()
```
Mixser, line 433: The weak reference pattern is defeated by acquiring `self` before the sleep. Consider moving `await asyncio.sleep()` before getting a strong ref to self. Or explain why we should keep the strong ref while sleeping

legion49f:
@Mixser my team is running into this issue and we are waiting for this PR to be merged. Would it be possible to make the changes you recommended and going ahead with a merge and a release? it's been longer than two months since the last time anyone commented on this thread. 

thanks!

kprzybyla:
I can make recommended changes if any maintainer of aiokafka is willing to review this. The comment by @Mixser is valid and I can fix this even today but I don't think that @Mixser will be able to push it further as I don't believe they are a maintainer of this project. We probably need someone like @ods here.

d-metrius:
Hello,

We also encounter issues with token re-authentication 1-2 times per day on average, and had to implement some ugly consumer restarts and producer retries in the code.
Are there any plans to merge this anytime soon? Would very much appreciate it!

Thanks!

legion49f:
We have been running this in production for over a year and it works. FYI

bwmetcalf:
We have a need for this as well.

chase-smithrx:
Any chance this could get merged in? We just encountered this in our production environment as well.

woz5999:
It's painful not having this. Maintainers, how can the community assist with moving this along?

ods:
@vmaurin Could you please take a look?

chase-smithrx:
I worked with Claude to vendor this change into our local setup. It made a few changes. I'll post a Claude generated writeup below in the hopes that this could be useful in getting this merged in:

> We ported this PR into an internal library rather than wait for the merge, since the
> hourly disconnects it fixes are very visible on Amazon MSK (SASL/OAUTHBEARER with IAM),
> where the broker advertises a ~1 hour `session_lifetime_ms`. **It is now running in
> production and it does what it says on the tin** — details below. The approach is right
> and the protocol support was already in place; thanks for writing it.
> 
> While building a test suite around the port, and then verifying it against a live
> cluster, we hit a handful of things that look worth addressing before this lands.
> Ordered roughly by severity.
> 
> ## 0. Production validation, for the merge-evidence pile
> 
> Since this is the thing a reviewer probably wants most: the port has now run overnight
> in production against Amazon MSK with SASL/OAUTHBEARER + IAM, on a consumer workload
> that previously logged the hourly disconnect ERRORs.
> 
> - The broker advertises a `session_lifetime_ms` of roughly **3,592,000** — just under the
>   hour, as expected for MSK. So the field is really populated on current brokers and this
>   PR's premise holds.
> - Re-authentication is scheduled at the jittered 85–95% mark and **fires and completes in
>   place**, on schedule, without dropping the connection.
> - The hourly `Failed fetch messages from` / `marking coordinator dead` /
>   `Heartbeat send request failed` ERROR bursts this was meant to remove are gone.
> 
> One caveat on how to read that: it exercises the **single-exchange** OAUTHBEARER path
> only. Every issue below that is specific to multi-exchange mechanisms (SCRAM, GSSAPI) is
> therefore *not* covered by this validation — it is exactly the region a live run like
> ours cannot reach.
> 
> Two of the fixes below were also in place for that run, so production never exercised
> the corresponding upstream paths. Where that is true I have said so rather than implying
> we observed the bug in the wild.
> 
> ## 1. A failed re-authentication can park every request on the connection forever
> 
> `_sasl_reauthentication` sets the gate only on the success path:
> 
> ```python
> await self._do_sasl_handshake()
> self._sasl_reauthentication_done.set()
> ```
> 
> and `_send_after_sasl_reauthentication` waits on it unbounded:
> 
> ```python
> await self._sasl_reauthentication_done.wait()
> return await self._send(request=request, expect_response=expect_response)
> ```
> 
> If `_do_sasl_handshake()` raises, `.set()` never runs. Most failure paths inside the
> handshake call `close()` first, which nulls `_writer` so subsequent `send()` calls raise
> early — but not all of them do. A `SaslAuthenticate` that simply times out propagates
> `asyncio.TimeoutError` out of `_do_sasl_handshake` **without** closing, and then:
> 
> - the gate stays clear forever;
> - the done-callback closes the connection with `AUTH_FAILURE`, which does not touch the
>   gate;
> - anything already parked in `_send_after_sasl_reauthentication` never wakes, and has no
>   timeout of its own, because the gate wait happens *before* `_send` wraps the future in
>   `wait_for(fut, self._request_timeout)`.
> 
> I modelled that control flow directly from the diff and the parked request never
> resolves:
> 
> ```
> handshake task done      : True
> connection closed with   : AUTH_FAILURE
> gate open                : False
> parked Fetch resolved    : False
> => Fetch still parked after an extra 0.5s with no timeout of its own
> ```
> 
> To be clear about provenance: this is modelled from the diff, not observed in the wild.
> Our production run shipped with both fixes below already applied, so it never exercised
> this path. I would still treat it as real — the reachability argument stands on its own.
> 
> Two independent fixes, both cheap, and worth doing together:
> 
> ```python
> self._sasl_reauthentication_done.clear()
> try:
>     await self._do_sasl_handshake()
> finally:
>     self._sasl_reauthentication_done.set()
> ```
> 
> and bounding the wait so a wedged re-authentication surfaces as a timeout the existing
> callers already handle, rather than as silence:
> 
> ```python
> async def _send_after_sasl_reauthentication(self, request, expect_response):
>     await asyncio.wait_for(
>         self._sasl_reauthentication_done.wait(), self._request_timeout
>     )
>     return await self._send(request=request, expect_response=expect_response)
> ```
> 
> ## 2. Re-authentication is scheduled per `SaslAuthenticate` exchange, orphaning tasks
> 
> The scheduling block sits inside the `while True:` authentication loop, right after
> `auth_bytes = resp.sasl_auth_bytes`. For a single-exchange mechanism such as
> OAUTHBEARER that runs once, so the bug is invisible there. Multi-exchange mechanisms run
> it more than once:
> 
> ```python
> self._sasl_reauthentication_task = self._create_sasl_reauthentication_task(
>     resp.session_lifetime_ms
> )
> ```
> 
> Each run **overwrites** the attribute without cancelling what was there, so every
> exchange but the last leaves an orphaned task that:
> 
> - `close()` cannot reach, since it only knows about the current attribute value;
> - fires a spurious re-authentication ~85–95% of a session lifetime later;
> - can close the connection from its own done-callback.
> 
> SCRAM-SHA-\* does two `SaslAuthenticate` round trips, and GSSAPI can do more, so this is
> reachable on real configurations rather than in theory.
> 
> The cleanest fix is to schedule once, after the loop, from the last response that
> carried a non-zero lifetime — a single re-authentication per handshake is the intent, and
> hoisting it also stops the behaviour depending on which exchanges the broker chooses to
> populate the field on. Failing that, cancel before replacing.
> 
> ## 3. If you adopt cancel-before-replace, exempt the running task
> 
> Worth flagging because we hit it in the port and it cost us a while to find. Scheduling
> is **reentrant**: the re-authentication task awaits `_do_sasl_handshake`, which sends a
> `SaslAuthenticate`, whose response schedules the next re-authentication — from inside
> the very task being replaced. A naive cancel-before-replace therefore cancels the
> running task, setting `_must_cancel`, and the next suspension point inside the handshake
> raises `CancelledError`.
> 
> That fails silently and in the worst possible way for a multi-exchange mechanism. Note
> `BaseSaslAuthenticator.step` dispatches through `run_in_executor`, so it *always*
> suspends:
> 
> - the remaining exchanges never run;
> - the done-callback returns early on `task.cancelled()`, so nothing closes and nothing
>   logs;
> - the gate reopens and a fresh timer is already armed;
> - the connection sits half-authenticated until the broker drops it at the session
>   deadline — the exact symptom this PR removes, minus the diagnostic.
> 
> OAUTHBEARER escapes only by accident, because `OAuthAuthenticator.step` returns `None`
> without awaiting on its second call and so never yields. Even there, every *successful*
> re-authentication ends up reporting as `cancelled()`, which is what hid it from our
> tests at first.
> 
> The guard is one clause:
> 
> ```python
> if task is not None and not task.done() and task is not asyncio.current_task():
>     task.cancel()
> ```
> 
> `asyncio.current_task()` raises `RuntimeError` rather than returning `None` when there is
> no running loop, and `close()` is reachable from `__del__`, so it needs wrapping if you
> call it from there.
> 
> Scheduling once after the loop (suggestion 2) sidesteps this entirely, which is another
> argument for that shape.
> 
> ## 4. The weakref is resolved before the sleep, so it holds a strong reference anyway
> 
> ```python
> self = self_ref()
> if self is None:
>     return
> await asyncio.sleep(sasl_reauthentication_time)
> ```
> 
> `self` is a strong reference held across the whole sleep — up to an hour on MSK — which
> defeats the `weakref` the surrounding code went to the trouble of taking. A connection
> dropped without an explicit `close()` stays alive until the timer fires, which is
> exactly the case `AIOKafkaConnection.__del__`'s unclosed-connection warning exists to
> catch.
> 
> Moving the resolution after the sleep restores the intent:
> 
> ```python
> await asyncio.sleep(sasl_reauthentication_time)
> 
> self = self_ref()
> if self is None:
>     return
> ```
> 
> ## 5. Two notes for the rebase
> 
> The `mergeable_state: dirty` is mostly mechanical, but two specifics may save some time:
> 
> - `SASL_REQUEST_API_KEYS` no longer builds. It relies on the request symbols being
>   iterable collections of versioned classes:
> 
>   ```python
>   frozenset(request.API_KEY for request in (*SaslHandShakeRequest, *SaslAuthenticateRequest))
>   ```
> 
>   On current `master` those are single `Request` classes, so this raises
>   `TypeError: Value after * must be an iterable, not ABCMeta`. It becomes
>   `frozenset({SaslHandShakeRequest.API_KEY, SaslAuthenticateRequest.API_KEY})`.
> 
> - The `send`/`_send` split is what actually conflicts, because it moves the body across
>   the `request.prepare(self._versions)` refactor. If you would rather not carry that
>   conflict again, the gate can be added at the top of the existing `send` without moving
>   anything, by having the deferred path re-enter `send` once the gate is open:
> 
>   ```python
>   async def _send_after_sasl_reauthentication(self, request, expect_response):
>       await asyncio.wait_for(
>           self._sasl_reauthentication_done.wait(), self._request_timeout
>       )
>       return await self.send(request, expect_response=expect_response)
>   ```
> 
>   I tried this against the unmodified `send` on 0.13.0 and it behaves: no recursion on
>   the open-gate path, exactly one write when the gate reopens, correct re-parking if a
>   second re-authentication starts while a request is waiting, and the `wait_for` still
>   bounds it. It does mean `send` returns a coroutine on the deferred path where it
>   otherwise returns an awaitable future — every in-tree caller awaits immediately so
>   that was fine, but it is your call whether that is a contract you want to widen.
> 
> ## 6. Worth a documentation note: KIP-368 requires a stable principal
> 
> This one bit us before any of the above, and it is invisible until you test against a
> real broker. KIP-368 only accepts a re-authentication that resolves to the **same
> principal** as the original authentication. For MSK IAM the principal includes the IAM
> role *session name*, and the usual AWS signer helper constructs a fresh botocore
> `Session` on every token it mints. Under IRSA, with `AWS_ROLE_SESSION_NAME` unset,
> botocore then generates a new timestamp-based role session name per session — so each
> handshake authenticates as a different principal and every re-authentication is
> rejected, showing up as `AUTH_FAILURE` closes rather than as anything that points at the
> cause.
> 
> The fix belongs in the caller's token provider, not here — resolve credentials once and
> reuse the resulting object, which keeps the fetcher and therefore the session name —
> but a sentence in the docs for this feature would save the next person the debugging.
> Something to the effect of: *a token provider must present a stable principal across
> tokens for re-authentication to be accepted; providers that construct fresh credentials
> per token will fail re-authentication.*
> 
> ## 7. In-flight requests during re-authentication
> 
> The gate stops *new* non-SASL sends but does not drain requests already awaiting
> responses, so a `SaslAuthenticate` can interleave with them on the wire. Kafka's Java
> client sidesteps this by re-authenticating only when the channel has no in-flight
> request.
> 
> I believe this is benign, because responses are correlated by `correlation_id` and the
> broker preserves per-connection response ordering. The overnight production run above is
> weak positive evidence: re-authentications interleaved with live consumer traffic and no
> protocol or correlation errors appeared. I would not call that a load test, though, and
> it is still a deliberate deviation from the Java behaviour — so it seems worth either a
> comment in the code or an explicit note in the PR that it was considered and accepted.
> 
> ## 8. Test coverage suggestion
> 
> If you add tests, a multi-exchange mechanism is where the value is. Suggestions 2 and 3
> are both invisible under OAUTHBEARER and PLAIN, and both are silent failures rather than
> exceptions, so a test that only asserts "a task was scheduled" passes with either bug
> present. What caught them for us was asserting the *terminal state* of the
> re-authentication task — `done() and not cancelled()` — and that every exchange in the
> handshake actually ran.

PR #1109, opened 2025-05-16
Remove typing_extensions from run time requirements

dbast:
### Changes

Remove typing_extensions from run time requirements = less dependencies to install for users of aiokafka.

### Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,7 @@ Improved Documentation:
 
 Misc:
 * Use SPDX license expression for project metadata.
+* Remove typing_extensions from run time requirements (pr #1109)
 
 
 0.12.0 (2024-10-26)
--- a/aiokafka/codec.py
+++ b/aiokafka/codec.py
@@ -1,9 +1,12 @@
+from __future__ import annotations
+
 import gzip
 import io
 import struct
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
-from typing_extensions import Buffer
+if TYPE_CHECKING:
+    from typing_extensions import Buffer
 
 _XERIAL_V1_HEADER = (-126, b"S", b"N", b"A", b"P", b"P", b"Y", 0, 1, 1)
 _XERIAL_V1_FORMAT = "bccccccBii"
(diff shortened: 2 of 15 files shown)


ods:
In general, it's better to avoid using `if TYPE_CHECKING` guards. They can be contagious (forcing library users to also use such guards) and may break code that evaluates annotations at run time (such as pydantic). However, starting with Python 3.12 `typing_extensions` is no longer necessary, and Python 3.11 has retired (in security-fix-only mode) for over a year. Therefore, it would be ideal to remove this requirement entirely for modern Python versions instead.

dbast:
@ods Thanks for the review. My thinking so far was:
* the `typing_extensions` package is not about specific Python versions. It brings new type system features from the latest stable Python version to older Python versions, as they claim it [here](https://github.com/python/typing_extensions?tab=readme-ov-file#overview). So the package itself won't go away if code bases make use of newly added features from it.
* aiokfaka has ruff + the pyupgrade rule activated ([see](https://github.com/aio-libs/aiokafka/blob/master/pyproject.toml#L91C14-L91C23)), which does [import replacements](https://github.com/asottile/pyupgrade?tab=readme-ov-file#import-replacements) like `from typing_extensions import Concatenate` -> `from typing import Concatenate`. So the usage of types from the  `typing_extensions` package are gradually phased out depending on the `requires-python` definition specified in the pyproject.toml. So `if TYPE_CHECKING` will cease existing in case everybody refrains from importing even newer type features from a newer `typing_extensions` package versions.
* `if TYPE_CHECKING` for me remains important for expensive imports for code that works with delayed imports for fast startups, to break circular imports and to not depend on packages with type definitions that are irrelevant during runtime as not evaluated.
* regarding pydantic: My understanding is that you can only be sure via [model_rebuild()](https://pydantic.com.cn/en/concepts/models/#rebuild-model-schema) that really all types are available at runtime ... types can be missing due to multiple reasons like `if TYPE_CHECKING` guarded imports but also "string-based type annotations". This code base does not use pydantic, so we do not have to make sure to run model_rebuild once against all models in unit tests to check that.

tl;dr so even if we introduce `if TYPE_CHECKING`, pyupgrade will phase it out in future when it becomes unneeded due newer min python version. And I personally see reasons to not depend on packages with type definitions that are irrelevant during runtime as not evaluated (there also is no pydantic in this codebase).

PR #1115, opened 2025-07-15
1115.bugfix: Make KafkaStorageError retriable after metadata refresh

omerhadari:
This error is supposed to be automatically retried by clients, but currently it is not.

This is how it is in the java implementation and in kafka-python

[Java Implementation](https://kafka.apache.org/30/javadoc/org/apache/kafka/common/errors/KafkaStorageException.html)
```
Client should request metadata update and retry if the response shows KafkaStorageException
```

[kafka-python](https://github.com/dpkp/kafka-python/blob/20e2d52ff2c337e02f8bac7af6c6e8d00ebcc63f/kafka/errors.py#L546)
```Python
class KafkaStorageError(BrokerResponseError):
    errno = 56
    message = 'KAFKA_STORAGE_ERROR'
    description = 'Disk error when trying to access log file on the disk.'
    retriable = True
    invalid_metadata = True
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,14 @@ Improved Documentation:
   (pr #1068 by @jzvandenoever)
 * Fix Java Client API reference (pr #1069 by @emmanuel-ferdman)
 
+
+
+Bugfixes:
+
+* Make KafkaStorageError retriable after metadata refresh like in other implementations (pr #1115 by @omerhadari)
+
+
+
 Misc:
 * Use SPDX license expression for project metadata.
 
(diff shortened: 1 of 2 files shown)


omerhadari:
Hi @ods, any chance this could get reviewed and a version released? Sorry for tagging you directly. We are getting errors in production which we have to bypass in ways that don't respect other client configs (e.g. request_timeout_ms).
Felt like this one is important because of the difference compared to the Java and other implementations.

Thank you!

ods:
Hi @omerhadari! Thanks for the contribution, looks good to me.

Issue #1085, opened 2025-01-14
[QUESTION] Support for Apache Kafka 4.0


ijuma:
Hi,

Apache Kafka 4.0 is removing a number of very old protocol api versions via [KIP-896](https://cwiki.apache.org/confluence/display/KAFKA/KIP-896%3A+Remove+old+client+protocol+API+versions+in+Kafka+4.0). 

When I was looking at the code, I noticed that we seem to use V0 or V1 below:

https://github.com/ulrikjohansson/aiokafka/blob/master/aiokafka/client.py#L179

Am I reading that right? If so, this will not work with Apache Kafka 4.0 and would have to change to use at least
Metadata request v4 (but ideally it would just use the newest version supported by the broker).

emasab:
I confirm that it's not working with AK 4.0 branch, the `simple_produce.py` example is giving this exception on the client:

```
Unable to request metadata from "localhost:42235": KafkaConnectionError: Connection at localhost:42235 closed
Traceback (most recent call last):
  File "/home/esabellico/workspace/aiokafka/examples/simple_produce.py", line 19, in <module>
    asyncio.run(send_one())
  File "/home/linuxbrew/.linuxbrew/opt/python@3.10/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/linuxbrew/.linuxbrew/opt/python@3.10/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/home/esabellico/workspace/aiokafka/examples/simple_produce.py", line 8, in send_one
    await producer.start()
  File "/home/esabellico/Standalone/venv3.10/lib/python3.10/site-packages/aiokafka/producer/producer.py", line 352, in start
    await self.client.bootstrap()
  File "/home/esabellico/Standalone/venv3.10/lib/python3.10/site-packages/aiokafka/client.py", line 266, in bootstrap
    raise KafkaConnectionError(f"Unable to bootstrap from {self.hosts}")
aiokafka.errors.KafkaConnectionError: KafkaConnectionError: Unable to bootstrap from [('localhost', 42235, <AddressFamily.AF_UNSPEC: 0>)]
Unclosed AIOKafkaProducer
producer: <aiokafka.producer.producer.AIOKafkaProducer object at 0x7ff528c80160>`
```
and it corresponds to this error log on the broker
```
[...]

PR #1136, opened 2025-11-18
Resolve API versions on connection

vmaurin:
As stated in the Kafka protocol documentation https://kafka.apache.org/protocol#api_versions the API versions should be solved for each single connection we open with a broker (the broker might be updated while we talk to it).

This commit aim to move all the versions logic into the protocol and the connection only instead of being spread in various classes.

It is introducing for that a new "Request" class that is acting as a builder for the Struct we are going to send into the wire. The Request class tries to collect all possible parameters, and then will react according the best available version in two possible ways:

* failing, as the intent of the caller was clearly to use a new feature that is not available on the broker yet
* best effort, as the caller can totally work with an older version of the API

This choice is made on per API basis.

From the caller perspective, if a different logic need to be ran according the API version, it can look up on the Response.API_VERSION

Doing such a change will help aiokafka supporting Kafka 4.0 as some API versions were removed as mentioned in #1085

It might not achieve the compatibility as we could lack some protocol definition, but it will definitively help.

### Changes

Fixes #

### Checklist

### DIFF
--- a/aiokafka/conn.py
+++ b/aiokafka/conn.py
@@ -427,8 +427,13 @@ class AIOKafkaConnection:
                 f"Connection at {self._host}:{self._port} broken: {err}"
             ) from err
 
-        if log.isEnabledFor(logging.DEBUG):
-            log.debug("%s Request %d: %s", repr(self), correlation_id, request_struct)
+        log.debug(
+            "Request to %s:%d %d: %s",
+            self._host,
+            self._port,
+            correlation_id,
+            request_struct,
+        )
 
         if not expect_response:
             return self._writer.drain()
@@ -464,7 +469,7 @@ class AIOKafkaConnection:
 
     def close(self, reason=None, exc=None):
         log.debug(
-            "Closing connection at %s:%s (%s,%s)", self._host, self._port, reason, exc
+            "Closing connection at %s:%s (%s, %s)", self._host, self._port, reason, exc
         )
         if self._reader is not None:
             self._writer.close()
@@ -488,7 +493,7 @@ class AIOKafkaConnection:
         if self._idle_handle is not None:
             self._idle_handle.cancel()
 
-        # transport.close() will close socket, but not right ahead
+        # transport.close() will close socket, but not right ahead.
         # Return a future in case we need to wait on it.
         return self._closed_fut
 
@@ -559,10 +564,13 @@ class AIOKafkaConnection:
 
             if not fut.done():
                 response = resp_type.decode(resp)
-                if log.isEnabledFor(logging.DEBUG):
-                    log.debug(
-                        "%s Response %d: %s", repr(self), correlation_id, response
-                    )
+                log.debug(
+                    "Response from %s:%d %d: %s",
+                    self._host,
+                    self._port,
+                    correlation_id,
+                    response,
+                )
                 fut.set_result(response)
 
         # Update idle timer.


aiokafka/admin/client.py
```diff
@@ -97,13 +94,13 @@ def __init__(
...
         security_protocol: str = "PLAINTEXT",
         ssl_context: SSLContext | None = None,
-        api_version: str = "auto",
         sasl_mechanism: str = "PLAIN",
         sasl_plain_username: str | None = None,
         sasl_plain_password: str | None = None,
         sasl_kerberos_service_name: str = "kafka",
         sasl_kerberos_domain_name: str | None = None,
         sasl_oauth_token_provider: str | None = None,
+        legacy_protocol: LegacyProtocol = False,
```
zladovan:
```suggestion
        legacy_protocol: LegacyProtocol | None = None,
```

As it's `None` as default in downstream. 
Also default in doc string need to be updated.

ijuma:
> * dropping support for old kafka (<0.11) will help remove a lot of code, maybe it is something to consider (I had to keep a configuration option a bit more explicit than "api_version" for that)

+1, it's not worth supporting <0.11 at this point.

vmaurin:
> > dropping support for old kafka (<0.11) will help remove a lot of code
> 
> I believe no one use such old version, so it would be better to drop support. @vmaurin, do you think it would be better/simpler to do this in a separate PR?

As you prefer @ods I am nearly done with the removal on top of the current PR : 
* pros: it is avoiding introducing this new settings "legacy_protocol", that is impacting quite a lot the code
* cons: it does two things at once

Personally I would do on this PR, so we don't need this new settings and all the related changes in `tests`

vmaurin:
@ods 0.9 and 0.10 support dropped as a separated commit in the same PR (so we avoid introducing "legacy_protocol" to then drop it)

aiokafka/producer/producer.py
```diff
@@ -350,25 +346,7 @@ async def start(self):
         ), "Please create objects with the same loop as running with"
         log.debug("Starting the Kafka producer")  # trace
         await self.client.bootstrap()
-
-        if self._compression_type == "lz4":
-            assert self.client.api_version >= (0, 8, 2), (
```
ods: What behaviour do we have in that case now?

aiokafka/protocol/api.py
```diff
@@ -59,32 +64,103 @@ class ResponseHeader_v1(Struct):
...
+    RESPONSE_TYPE: ClassVar[type[Response]]
+    SCHEMA: ClassVar[Schema]
+
+    def __init_subclass__(cls) -> None:
+        super().__init_subclass__()
+        if (
+            cls.API_KEY is None
+            or cls.API_VERSION is None
+            or cls.RESPONSE_TYPE is None
+            or cls.SCHEMA is None
```
ods: If we forget to define some of them, they still won't be None. Probably we don't need those checks at all?

aiokafka/protocol/commit.py
```diff
@@ -244,69 +263,36 @@ class OffsetFetchRequest_v2(Request):
...
+
+    def __init__(
+        self, consumer_group: str, partitions: list[tuple[str, list[int]]] | None
+    ):
+        self._consumer_group = consumer_group
+        self._partitions = partitions
+
+    def build(self, request_struct_class: type[RequestStruct]) -> RequestStruct:
+        if request_struct_class.API_VERSION < 2 and self._partitions is None:
+            raise IncompatibleBrokerVersion(
```
ods: Looks like the exception changed for this case. This one looks more appropriate to me, but I'm not sure if we don't need to mention about it in the change log.

ods (approved):
There are some minor remarks, but in general I see no blockers to merge it.

ijuma:
> So, we can safely stop using old formats in producer, but it would be better to support them in consumer when they are returned by broker. Right?

Correct.

aiokafka/protocol/commit.py
```diff
@@ -116,29 +118,46 @@ class OffsetCommitRequest_v2(Request):
...
+        topics: list[tuple[str, list[tuple[int, int, str]]]],
+    ):
+        self._consumer_group = consumer_group
+        self._consumer_group_generation_id = consumer_group_generation_id
+        self._consumer_id = consumer_id
+        self._retention_time = retention_time
+        self._topics = topics
+
+    def build(self, request_struct_class: type[RequestStruct]) -> RequestStruct:
+        return request_struct_class(
```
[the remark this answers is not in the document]
vmaurin: True. We can probably create a type parameter here, that could be a union of all "CLASSES" (could it replace the CLASSES class var then ?) and eventually replace the conditional on the API_VERSION with some pattern matching on the type ?

aiokafka/producer/producer.py
vmaurin: Now, it will fails later in the process, when trying to actually send a query to the producer: 
* either the request class is raising a IncompatibleBroker exception, because it knows already that the provided parameters are not compatible with the negotiated version
* or the request is actually sent to the broker, that will return an error

aiokafka/protocol/api.py
vmaurin: There is a test in test_requests.py and it will raise an AttributeError. Maybe I could do a better testing to raise a single type of error (like "the attribute need to be define + being not None and the proper type" ?)

tests/test_transactional_producer.py
```diff
@@ -24,7 +24,7 @@ async def test_producer_transactions_not_supported(self):
             bootstrap_servers=self.hosts,
             transactional_id="sobaka_producer",
         )
-        with self.assertRaises(UnsupportedVersionError):
+        with self.assertRaises(KafkaError):
```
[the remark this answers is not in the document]
vmaurin: I think I had different error with 0.9 and 0.10. Since we stopped the support of these, I think I can just drop this test

aiokafka/protocol/commit.py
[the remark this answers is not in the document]
vmaurin: I have pushed a new commit doing something like above

vmaurin:
> @vmaurin @ijuma Are there any other questions, or shall I proceed with the merge?

I am still a bit worried by the failing test, but I just can't reproduce it locally. I am seen something weird with the writer drain https://github.com/aio-libs/aiokafka/discussions/1138 but it is probably not related

vmaurin:
> @vmaurin I see one test became flaky with "BrokenPipeError: [Errno 32] Broken pipe" error after this change, while it succeeds for other your branch. Any idea why it may happen?

I have a lead @ods . I fix a garbage collection issue of the connection being passed to log.debug that was preventing the proper collection. But as it is not held by the log, it is probably getting collected to early here. Let me see if I can find a fix

vmaurin:
For the current flaky tests, it has nothing to do with garbage collection. I have dumped the server logs and we do have an issue
```
INFO [KafkaApi-0] Closing connection due to error during produce request with correlation id 2 from client id aiokafka with ack=0
Topic and partition to exceptions: b'foo'-0 -> org.apache.kafka.common.errors.UnknownTopicOrPartitionException (kafka.server.KafkaApis)
```

So the server is closing the connection, resulting in the broken pipe. In my local, it is not fast enough to close the connection before the test end

aiokafka/conn.py
```diff
@@ -530,8 +488,8 @@ def close(self, reason=None, exc=None):
         if self._idle_handle is not None:
             self._idle_handle.cancel()
 
-        # transport.close() will close socket, but not right ahead. Return
-        # a future in case we need to wait on it.
+        # transport.close() will close socket, but not right ahead
+        # Return a future in case we need to wait on it.
```
ods: _nit:_
```suggestion
        # transport.close() will close socket, but not right ahead.
        # Return a future in case we need to wait on it.
```

vmaurin:
> Thanks! Great work. Let me know when you think we’re ready to merge.

I guess it is ready. I did a sanity check on my test project (so running with producer/consumer/transactional loop/admin on a cluster compose of 3 workers) and I didn't see errors

PR #1139, opened 2025-11-27
Simplify flexible versions

vmaurin:
The flexible versions is a protocol specificity for newer versions of the API. When an API is flexible, it is using more compact structures and also allow additional "dynamic" fields that could be added without the need to introduce a new API versions.

This commit move the flexible versions support to the protocol layer, so it is more transparent and easy when defining Struct classes and schemas.

When defining the schema, we can specify a tagged field with a tuple containing the field name and the field tag.

### Checklist

### DIFF
--- a/aiokafka/protocol/abstract.py
+++ b/aiokafka/protocol/abstract.py
@@ -8,11 +8,11 @@ T = TypeVar("T")
 class AbstractType(Generic[T], metaclass=abc.ABCMeta):
     @classmethod
     @abc.abstractmethod
-    def encode(cls, value: T) -> bytes: ...
+    def encode(cls, value: T, flexible: bool) -> bytes: ...
 
     @classmethod
     @abc.abstractmethod
-    def decode(cls, data: BytesIO) -> T: ...
+    def decode(cls, data: BytesIO, flexible: bool) -> T: ...
 
     @classmethod
     def repr(cls, value: T) -> str:
--- a/aiokafka/protocol/types.py
+++ b/aiokafka/protocol/types.py
@@ -19,6 +19,10 @@ T = TypeVar("T")
 
 ValueT: TypeAlias = Union[type[AbstractType[Any]], "String", "Array", "Schema"]
 
+TaggedFieldId: TypeAlias = tuple[str, int]
+
+FieldId: TypeAlias = TaggedFieldId | str
+
 
 def _pack(f: Callable[[T], bytes], value: T) -> bytes:
     try:
@@ -47,11 +51,11 @@ class Int8(AbstractType[int]):
     _unpack = struct.Struct(">b").unpack
 
     @classmethod
-    def encode(cls, value: int) -> bytes:
+    def encode(cls, value: int, flexible: bool = False) -> bytes:
         return _pack(cls._pack, value)
 
     @classmethod
-    def decode(cls, data: BytesIO) -> int:
+    def decode(cls, data: BytesIO, flexible: bool = False) -> int:
         return _unpack(cls._unpack, data.read(1))
 
 
@@ -60,11 +64,11 @@ class Int16(AbstractType[int]):
     _unpack = struct.Struct(">h").unpack
 
     @classmethod
-    def encode(cls, value: int) -> bytes:
+    def encode(cls, value: int, flexible: bool = False) -> bytes:
         return _pack(cls._pack, value)
 
     @classmethod
-    def decode(cls, data: BytesIO) -> int:
+    def decode(cls, data: BytesIO, flexible: bool = False) -> int:
         return _unpack(cls._unpack, data.read(2))
 
 
@@ -73,11 +77,11 @@ class Int32(AbstractType[int]):
     _unpack = struct.Struct(">i").unpack
 
     @classmethod
-    def encode(cls, value: int) -> bytes:
+    def encode(cls, value: int, flexible: bool = False) -> bytes:
         return _pack(cls._pack, value)
 
     @classmethod
-    def decode(cls, data: BytesIO) -> int:
+    def decode(cls, data: BytesIO, flexible: bool = False) -> int:
         return _unpack(cls._unpack, data.read(4))
 
 
@@ -86,11 +90,11 @@ class UInt32(AbstractType[int]):
     _unpack = struct.Struct(">I").unpack
 
     @classmethod
-    def encode(cls, value: int) -> bytes:
+    def encode(cls, value: int, flexible: bool = False) -> bytes:
         return _pack(cls._pack, value)
 
     @classmethod
-    def decode(cls, data: BytesIO) -> int:
+    def decode(cls, data: BytesIO, flexible: bool = False) -> int:
         return _unpack(cls._unpack, data.read(4))
 
 
@@ -99,11 +103,11 @@ class Int64(AbstractType[int]):
     _unpack = struct.Struct(">q").unpack
 
     @classmethod
-    def encode(cls, value: int) -> bytes:
+    def encode(cls, value: int, flexible: bool = False) -> bytes:
         return _pack(cls._pack, value)
 
     @classmethod
-    def decode(cls, data: BytesIO) -> int:
+    def decode(cls, data: BytesIO, flexible: bool = False) -> int:
         return _unpack(cls._unpack, data.read(8))
 
 
@@ -112,26 +116,39 @@ class Float64(AbstractType[float]):
     _unpack = struct.Struct(">d").unpack
 
     @classmethod
-    def encode(cls, value: float) -> bytes:
+    def encode(cls, value: float, flexible: bool = False) -> bytes:
         return _pack(cls._pack, value)
 
     @classmethod
-    def decode(cls, data: BytesIO) -> float:
+    def decode(cls, data: BytesIO, flexible: bool = False) -> float:
         return _unpack(cls._unpack, data.read(8))
 
 
 class String:
-    def __init__(self, encoding: str = "utf-8"):
+    def __init__(self, encoding: str = "utf-8", allow_flexible: bool = True):
         self.encoding = encoding
+        self.allow_flexible = allow_flexible
 
-    def encode(self, value: str | None) -> bytes:
+    def encode(self, value: str | None, flexible: bool) -> bytes:
         if value is None:
-            return Int16.encode(-1)
+            return (
+                UnsignedVarInt32.encode(0)
+                if flexible and self.allow_flexible
+                else Int16.encode(-1, flexible)
+            )
         encoded_value = str(value).encode(self.encoding)
-        return Int16.encode(len(encoded_value)) + encoded_value
+        return (
+            UnsignedVarInt32.encode(len(encoded_value) + 1) + encoded_value
+            if flexible and self.allow_flexible
+            else Int16.encode(len(encoded_value), flexible) + encoded_value
+        )
 
-    def decode(self, data: BytesIO) -> str | None:
-        length = Int16.decode(data)
+    def decode(self, data: BytesIO, flexible: bool) -> str | None:
+        length = (
+            UnsignedVarInt32.decode(data) - 1
+            if flexible and self.allow_flexible
+            else Int16.decode(data, flexible)
+        )
         if length < 0:
             return None
         value = data.read(length)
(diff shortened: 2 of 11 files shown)


vmaurin:
@ods A small reminder about this one, let me know if I should make it ready to merge again (it is a bit outdated)

> About this MR: There is already a flexible field implementations in the current master branch, but it has a bug + it is not very convenient to define the message schemas. My idea with this MR is to make the flexible fields easy to define and use, similar to what it is done in the official java client json "schemas".

aiokafka/protocol/types.py
```diff
@@ -174,33 +201,94 @@ class Boolean(AbstractType[bool]):
...
+                        result[i] = self.fields[i].decode(
+                            BytesIO(encoded_value), flexible
+                        )
+
+        return tuple(result)
+
+    @staticmethod
+    def _encode_tagged_fields(value: dict[int, bytes]) -> bytes:
+        ret = UnsignedVarInt32.encode(len(value))
+        for k, v in value.items():
```
GlebShipilov, line 273: KIP-482 requires tagged fields on the wire in strictly ascending tag order. value.items() iterates in insertion (declaration) order, not tag order. If someone declares (("b", 5), Int16) before (("a", 2), Int16), the broker can reject the request.

Fix: for k, v in sorted(value.items()):

aiokafka/protocol/types.py
```diff
@@ -174,33 +201,94 @@ class Boolean(AbstractType[bool]):
...
+        return b"".join(
+            field.encode(item[i], flexible)
+            for i, field in enumerate(self.fields)
+            if self.tags[i] is None
+        ) + (
+            self._encode_tagged_fields(
+                {
+                    self.tags[i]: field.encode(item[i], flexible)
+                    for i, field in enumerate(self.fields)
+                    if self.tags[i] is not None
```
GlebShipilov, line 244: When a tagged field's value is None (the default from Struct.__init__), Int16.encode(None) will crash with struct.error. For String/Bytes, None encodes as a null marker, but per spec a tagged field at its default value should be omitted from the wire entirely, not sent as null.
Minimal fix: add and item[i] is not None to the comprehension filter:
`> if self.tags[i] is not None and item[i] is not None`

tests/test_protocol_object_conversion.py
```diff
@@ -188,6 +190,53 @@ def test_with_complex_nested_array(
         assert myarray[1]["subarray"][0]["innertest"] == "hello"
         assert myarray[1]["subarray"][0]["otherinnertest"] == "hello again"
 
+    def test_flexible_version(self, superclass: type[RequestStruct | Response]) -> None:
```
GlebShipilov, line 193: This test only does a happy-path round-trip with all tagged fields populated. 
Please add cases for:

- Decode a payload that lacks a known tagged field - the schema slot should stay None.
- Decode a payload with an unknown tag id - it should be silently skipped, and subsequent fields should still decode correctly.
- Encode with a tagged field set to None - it should be omitted from the wire (see comment on types.py).

aiokafka/protocol/types.py
```diff
@@ -293,128 +397,3 @@ def encode(cls, value: int) -> bytes:
...
-        ret = b""
-        while (v & 0xFFFFFFFFFFFFFF80) != 0:
-            b = (value & 0x7F) | 0x80
-            ret += struct.pack("B", b)
-            v >>= 7
-        ret += struct.pack("B", v)
-        return ret
-
-
-class CompactString(String):
```
GlebShipilov, line 341: CompactString, CompactArray, CompactBytes, and TaggedFields were importable from aiokafka.protocol.types. Downstream code using them will break with no migration path.

Consider keeping one-line compat shims with a DeprecationWarning:
`
> CompactString = String  # deprecated, use String + FLEXIBLE_VERSION
> CompactBytes = Bytes
> CompactArray = Array
`

At minimum, call out the removed symbols by name in the changelog.

aiokafka/protocol/abstract.py
```diff
@@ -8,11 +8,11 @@
 class AbstractType(Generic[T], metaclass=abc.ABCMeta):
     @classmethod
     @abc.abstractmethod
-    def encode(cls, value: T) -> bytes: ...
+    def encode(cls, value: T, flexible: bool) -> bytes: ...
```
GlebShipilov, line 11: Int8/Int16/Int32/Int64 default flexible=False, but Bytes, Boolean, and String (via the abstract base) now require it as a mandatory positional. This is inconsistent, any downstream call like Bytes.encode(value) that existed before this PR will TypeError. Either default flexible=False everywhere or nowhere.

tests/test_protocol.py
```diff
@@ -328,33 +327,38 @@ def test_unsigned_varint_serde() -> None:
 
 
 def test_compact_data_structs() -> None:
```
GlebShipilov, line 329: nit: The concept is no longer called "compact" in this PR, it's "flexible". Consider renaming to test_flexible_serde so the test name matches the new terminology.

aiokafka/protocol/types.py
```diff
@@ -174,33 +201,94 @@ class Boolean(AbstractType[bool]):
...
-    def __init__(self, *fields: tuple[str, ValueT]):
+    def __init__(self, *fields: tuple[FieldId, ValueT]):
         if fields:
-            self.names, self.fields = zip(*fields, strict=False)
+            tagged_names, values = zip(
+                *(
+                    (key, value) if isinstance(key, tuple) else ((key, None), value)
+                    for key, value in fields
+                ),
+                strict=False,
```
GlebShipilov, line 224: There's no reason for the lengths to disagree here, strict=True would be safer and document intent

GlebShipilov (commented):
Three potentially blocking issues I found:

1. Tagged fields must be sorted by tag on encode (KIP-482).
2. None-valued tagged fields must be omitted, not serialized (crashes for Int types, wrong semantics for String/Bytes).
3. Need tests for decode with missing/unknown tagged fields — the forward-compat path that makes this whole feature worthwhile.

Plus: public symbols removed without deprecation shim, and inconsistent flexible defaults across type classes. See inline comments.

aiokafka/protocol/types.py
vmaurin: Right, I missed that from the KIP

vmaurin:
Thank you for all the very valuable feedback @GlebShipilov I will push updates soon

aiokafka/protocol/abstract.py
vmaurin: I am not so happy about this boolean to be honest. In java client, they do have still two separated classes (like String and CompactString, and even the nullable variant here) in the code base, but then they are used only for serialization, not for defining the schema (that are achieve with a json file + some code generation)

I think it is the right approach so far of aiokafka to define schemas as python, but maybe we should not use the same set of classes for definition and serialization ?

vmaurin: For example, here a piece of the generate java code from their json

```
    public static final Schema SCHEMA_18 =
        new Schema(
            new Field("max_wait_ms", Type.INT32, "The maximum time in milliseconds to wait for the response."),
            new Field("min_bytes", Type.INT32, "The minimum bytes to accumulate in the response."),
            new Field("max_bytes", Type.INT32, "The maximum bytes to fetch.  See KIP-74 for cases where this limit may not be honored."),
            new Field("isolation_level", Type.INT8, "This setting controls the visibility of transactional records. Using READ_UNCOMMITTED (isolation_level = 0) makes all records visible. With READ_COMMITTED (isolation_level = 1), non-transactional and COMMITTED transactional records are visible. To be more concrete, READ_COMMITTED returns all data from offsets smaller than the current LSO (last stable offset), and enables the inclusion of the list of aborted transactions in the result, which allows consumers to discard ABORTED transactional records."),
            new Field("session_id", Type.INT32, "The fetch session ID."),
            new Field("session_epoch", Type.INT32, "The fetch session epoch, which is used for ordering requests in a session."),
            new Field("topics", new CompactArrayOf(FetchTopic.SCHEMA_18), "The topics to fetch."),
            new Field("forgotten_topics_data", new CompactArrayOf(ForgottenTopic.SCHEMA_13), "In an incremental fetch request, the partitions to remove."),
            new Field("rack_id", Type.COMPACT_STRING, "Rack ID of the consumer making this request."),
            TaggedFieldsSection.of(
                0, new Field("cluster_id", Type.COMPACT_NULLABLE_STRING, "The clusterId if known. This is used to validate metadata fetches prior to broker registration."),
                1, new Field("replica_state", ReplicaState.SCHEMA_15, "The state of the replica in the follower.")
            )
        );
```

@ods @GlebShipilov what would be for you the ideal solutions:
* single classes (String/Int32/Array) with a boolean (like in this PR)
* many classes, explicit "TaggedFields" property named "tag" (like in master branch)
* many classes, a TaggedFieldsSection thing we can add in the schema (similar to the java generated classes)
* a code generator from the same json as java client

What I don't like so much from the master branch is:
* introduce a field name "tags", that then a bit weird to use (ex `response.tags["something"]`)
* when it is a flexible version, we need to be sure to use Compact variant everywhere
* when it is a flexible version, we need to be sure that all schemas/sub schemas would contains TaggedFields

ods: > I think it is the right approach so far of aiokafka to define schemas as python, but maybe we should not use the same set of classes for definition and serialization ?

I don't mind having separate classes for definition and serialization. But I don't have full picture yet how it may look like. Maybe it's worth to fork this this branch and try?

GlebShipilov: > I think it is the right approach so far of aiokafka to define schemas as python, but maybe we should not use the same set of classes for definition and serialization? - @vmaurin

> I don't mind having separate classes for definition and serialization. But I don't have full picture yet how it may look like. Maybe it's worth to fork this branch and try? - @ods

I agree with both of you, and I think the discomfort we all feel about the flexible: bool parameter is a symptom, not the disease. The real problem is that one set of classes is being asked to play two roles:
1. schema definition (what fields exist, what their semantics are) 
2. wire codec (how those fields look on the network in a given protocol generation).  
Those two concerns have different lifecycles: the schema is fixed at class-definition time, but the wire format depends on the API version. Squeezing both into a single class forces us to thread a runtime flag through every encode/decode call, which is exactly how we ended up with inconsistent defaults, the allow_flexible=False escape hatch on String, and the Int*.encode(None) crash for default tagged fields.

Of the four options @vmaurin laid out, here's how I'd rank them:

1. TaggedFieldsSection + explicit Compact* types in the schema (Java-style, no codegen). This is my preferred next step. Schemas would look like:

```
SCHEMA = Schema(
    ("max_wait_ms", Int32),
    ("rack_id", CompactString()),
    TaggedFieldsSection(
        (0, "cluster_id", CompactNullableString()),
        (1, "replica_state", ReplicaState.SCHEMA_15),
    ),
)
```
Why I like it:
- No flexible: bool anywhere. String and CompactString are different classes that simply know how to serialize themselves. The boolean stops propagating virally through every codec method.
- FLEXIBLE_VERSION becomes redundant, the presence of TaggedFieldsSection in the schema is the flexibility marker. Single source of truth.
- Tagged fields stop being "regular fields with a tag" and become what they actually are at the wire level: a dedicated trailing section. This is a more honest model and eliminates my complaint about response.tags["something"], tagged fields can be exposed as normal attributes on the Struct, while the section object owns the encode/decode and ordering rules (sorting by tag, omitting defaults, skipping unknown tags on decode).
- It's structurally impossible to forget the tagged-fields section in a flexible schema, __init_subclass__ can validate it.
The migration is large but mechanical, easy to review schema-by-schema.

2. Code generation from the upstream Kafka JSON message specs (option 4). Strategically this is the right answer, because it's how the Java client, Sarama, librdkafka, kafka-go and most production clients stay in sync with the protocol, and it would solve correctness issues like KIP-482 ordering exactly once in a template. But it's a multi-week investment (writing the generator, vendoring/syncing the JSON specs, integrating into CI, etc..), and I don't think it should block this PR. It's a great long-term goal.

3. The current PR's single-class + boolean approach. Acceptable as a transitional state if we fix the obvious blockers (KIP-482 sort, omit None tagged fields, add the missing forward-compat decode tests, and either default flexible=False everywhere or restore deprecation shims for Compact*/TaggedFields/VarInt* so we don't break downstream imports). But I would explicitly mark it as "interim implementation, see #XXXX for redesign" in CHANGES, otherwise it tends to ossify.

4. Master's tags field per Struct. I agree with @vmaurin's objections to it: response.tags[...] is awkward, Compact* has to be remembered everywhere by hand, and it's easy to forget TaggedFields on a sub-schema. Going back is not the right move.

My proposal so we don't deadlock this PR on architecture:

1. Land PR with the blocker fixes (sort, None-omission, shims, consistent default) as the minimum correct flexible-versions implementation. It's strictly better than master and unblocks people who need flexible APIs now.
2. Open a follow-up issue "Restructure protocol schema definition (post-KIP-482)" with the four options above and a vote.
3. To make the conversation concrete, pick a small canary API (e.g. AlterPartitionReassignmentsRequest/Response_v0 is a good one (it has tagged fields, compact strings, compact arrays, and nested structs, all in a low-traffic API)) and prototype option 1 against it on a fork. Once we see the resulting schema and codec code side-by-side with the current shape, the choice between options 1 and 4 (and the maintenance cost of each) will be much easier to evaluate.

vmaurin: Following @GlebShipilov advice, here a revisited PR https://github.com/aio-libs/aiokafka/pull/1162

vmaurin:
Closing this one in favor of https://github.com/aio-libs/aiokafka/pull/1162 instead (that is a smaller step going in a better direction)

PR #1142, opened 2025-12-09
Use linger_ms to delay batches

### LINKED ISSUE #1137: Linger time acting like a rate limit, when its not supposed to do that
**Describe the bug**
When a batch can not fit a message, kafka waits until the current batch is delivered to add the message to the next batch. Linger time prevents batches from being delivered until a cooldown period has expired. Combined this makes it so that large messages time out.

**Expected behaviour**
From what I understand about linger time, its meant to set a maximum wait time until the message is to be delivered to message broker. Its not meant to act as a rate limit, but instead a mechanism to ensure a few large packets are sent out rather than many small ones. 

 - aiokafka version (`python -c "import aiokafka; print(aiokafka.__version__)"`): 0.12.0
 - Kafka Broker version (`kafka-topics.sh --version`): 4.0.0

**Reproducible example**

I used the files in [this gist](https://gist.github.com/Ktmi/9271edecd5df07af778404744d056781) for looking into the problem. The python program was for testing sending many large messages, the json file was the data we were sending, and the patch was something I created to look at how `aiokafka` was sending it.

vmaurin:
Currently, the linger_ms parameter is used to limit the number of
request sent to a given broker as adding at minimum linger_ms between to
produce requests. It doesn't produce the expected result when the max
size of batches is also reached.

The purpose of linger_ms should be not limiting the rate of request, but
instead keeping a batch still open long enough so it has the opportunity
to grow bigger.

Moving linger_ms to the message accumulator (like in the java client)
gives the opportunity to have a condition both on the max size and the
lingering time.

fixes #1137

### Checklist

### DIFF
--- a/aiokafka/producer/message_accumulator.py
+++ b/aiokafka/producer/message_accumulator.py
@@ -1,6 +1,5 @@
 import asyncio
 import collections
-import contextlib
 import copy
 import time
 from collections.abc import Sequence
@@ -137,12 +136,11 @@ class BatchBuilder:
 class MessageBatch:
     """This class encapsulate operations with batch of produce messages"""
 
-    def __init__(self, tp, builder, ttl, linger_time, max_size):
+    def __init__(self, tp, builder, ttl, linger_time):
         self._builder = builder
         self._tp = tp
         self._ttl = ttl
         self._linger_time = linger_time
-        self._max_size = max_size
         self._ctime = time.monotonic()
 
         # Waiters
@@ -341,7 +339,7 @@ class MessageAccumulator:
         self._compression_type = compression_type
         self._batch_ttl = batch_ttl
         self._waiter_future = loop.create_future()
-        self._wakeup_task = None
+        self._wakeup_handle = None
         self._closed = False
         self._txn_manager = txn_manager
         self._linger_time = linger_ms / 1000
@@ -469,7 +467,7 @@ class MessageAccumulator:
         self._pending_batches.remove(batch)
         batch.reset_drain()
 
-    async def drain_by_nodes(self, ignore_nodes, muted_partitions=frozenset()):
+    def drain_by_nodes(self, ignore_nodes, muted_partitions=frozenset()):
         """Group batches by leader to partition nodes."""
         nodes = collections.defaultdict(dict)
         unknown_leaders_exist = False
@@ -523,23 +521,20 @@ class MessageAccumulator:
         self._waiter_future = self._loop.create_future()
 
         # cleaning up old wakeup task
-        if self._wakeup_task:
-            self._wakeup_task.cancel()
-            with contextlib.suppress(asyncio.CancelledError):
-                await self._wakeup_task
-            self._wakeup_task = None
+        if self._wakeup_handle:
+            self._wakeup_handle.cancel()
+            self._wakeup_handle = None
 
         # schedule a new wakeup task
         if remaining_linger_time:
-            self._wakeup_task = self._loop.create_task(
-                self._wakeup(self._waiter_future, remaining_linger_time)
+            self._wakeup_handle = self._loop.call_later(
+                remaining_linger_time, self._wakeup, self._waiter_future
             )
 
         return nodes, unknown_leaders_exist
 
     @staticmethod
-    async def _wakeup(fut, after):
-        await asyncio.sleep(after)
+    def _wakeup(fut):
         if not fut.done():
             fut.set_result(None)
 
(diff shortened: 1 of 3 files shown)


aiokafka/producer/message_accumulator.py
```diff
@@ -266,6 +273,19 @@ async def wait_drain(self, timeout=None):
...
+        """Return the eventual remaining linger time"""
+        lifetime = time.monotonic() - self._ctime
+        # Batch builders reject a message if they reach the exact size
+        # so we should consider it ready if it is size - 1
+        if (
+            self._builder.closed()
+            or self._builder.size() >= self._max_size - 1
+            or lifetime >= self._linger_time
+        ):
+            return None
```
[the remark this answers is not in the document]
vmaurin: I would tend to keep "None" for no lingers, as returning 0 for me would mean "wake up immediately", but here that is not the meaning. When there are no batch lingering, we don't wait for a given time. The sender routine will be awaken only if new data shows up, or a transaction is closed.

Also in practice, it is about to change this return, and eventually the initialization of the `remaining_linger_time` but we have to keep the conditional logic as it is, so it doesn't simplify anything in `drain_by_nodes`

aiokafka/producer/message_accumulator.py
```diff
@@ -396,18 +422,22 @@ async def add_message(
             if future is not None:
                 return future
             # Batch is full, can't append data atm,
-            # waiting until batch per topic-partition is drained
+            # wake up the sender loop
+            # and wait until batch per topic-partition is drained
+            if not self._waiter_future.done():
+                self._waiter_future.set_result(None)
```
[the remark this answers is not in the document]
vmaurin: It is fixed now, by closing a batch builder when it is full and not relying on the size in the linger method

vmaurin:
> @ods We are running this one in prod for several weeks so it is ready to review/merge I would say [master...vmaurin:aiokafka:pre-release](https://github.com/aio-libs/aiokafka/compare/master...vmaurin:aiokafka:pre-release)

@ods I resolved the conflict here, so it is ready to be reviewed (and eventually merged ?). On our setup, in production, we can leverage this lingering behavior now and it did allow us to decrease the number of message sent to our brokers

aiokafka/producer/message_accumulator.py
```diff
@@ -478,14 +515,33 @@ def drain_by_nodes(self, ignore_nodes, muted_partitions=frozenset()):
...
-            self._wait_data_future.set_result(None)
-        self._wait_data_future = self._loop.create_future()
+        if not self._waiter_future.done():
+            self._waiter_future.set_result(None)
+        self._waiter_future = self._loop.create_future()
+
+        # cleaning up old wakeup task
+        if self._wakeup_task:
+            self._wakeup_task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
```
ods: If the task calling drain_by_nodes is cancelled at this moment, then the CancelledError exception will be swallowed and execution will continue effectively ignoring cancel. This doesn't look like correct behaivior to me.

aiokafka/producer/message_accumulator.py
```diff
@@ -442,10 +468,11 @@ def reenqueue(self, batch):
         self._pending_batches.remove(batch)
         batch.reset_drain()
 
-    def drain_by_nodes(self, ignore_nodes, muted_partitions=frozenset()):
+    async def drain_by_nodes(self, ignore_nodes, muted_partitions=frozenset()):
```
ods: Do I understand correctly, that with `call_later` instead of task with `sleep` we can leave it non-async?

vmaurin: I moved it async to do the cleanup of the existing wake up task before we schedule a new one. The current task with sleep is anyway keeping a future reference to the task instead of relying on "self._waiter_future". I guess I can use a call_later / handle.cancel instead

aiokafka/producer/message_accumulator.py
vmaurin: That is right, could be removed then with a call_later. We probably have the same issue in a similar pattern in aiokafka/consumer/fetcher.py

aiokafka/producer/message_accumulator.py
vmaurin: @ods updated with a call_later

aiokafka/producer/message_accumulator.py
vmaurin: Removed

ods:
LGTM. Have you already tried this version version in production? If so, I'd like to merge it.

vmaurin:
> LGTM. Have you already tried this version version in production? If so, I'd like to merge it.

I had the version with the async/await. For the change with call_later, I only did a sanity check on a test project (still 3 brokers/many producers/consumers pattern)

PR #1150, opened 2026-01-31
Return and deprecate api_version (fixes #1147)

### LINKED ISSUE #1147: Revert removal of api_version from AIOKafkaProducer/Consumer.__init__ or make it deprecated with a warning
Hello aiokafka team,

After updating to the latest release, our application started failing with the error:

```
AIOKafkaProducer.__init__() got an unexpected keyword argument 'api_version'
```

We traced the change to [PR #1136](https://github.com/aio-libs/aiokafka/pull/1136), where the api_version parameter was removed from the __init__ constructors of AIOKafkaProducer and AIOKafkaConsumer, and the logic was moved into a separate AIOKafkaConfig class.

While we understand the motivation behind the refactoring to improve configuration management, the sudden and breaking removal of a public constructor parameter in a minor release (if it wasn't a major version bump) causes significant disruption for downstream users.

**The Problem:**

**Breaking Change:** The api_version parameter was a documented and widely used part of the public API. Its immediate removal breaks existing codebases that instantiate clients directly, without any deprecation period.
**Upgrade Friction:** This forces all users to immediately refactor their code, even for simple applications. It's especially problematic in larger codebases where Kafka client initialization might be scattered across multiple modules.
**Inconsistency with kafka-python:** aiokafka aims to be compatible with kafka-python. The synchronous KafkaProducer/KafkaConsumer from kafka-python still accepts api_version directly in its constructor, making this change a divergence that hurts migration and familiarity.
**Our Request / Proposal:**
We kindly ask you to consider one of the following solutions:

**Preferred Solution:** Re-introduce the api_version parameter to __init__, but mark it as deprecated with a DeprecationWarning. The parameter can be internally mapped to the new AIOKafkaConfig object. This provides a smooth migration path:

Existing code continues to work without crashing.
Users get a clear warning pointing them to the new configuration method.
You can schedule the parameter for removal in a future major release (e.g., 2.0).
Alternative: If re-adding the parameter is absolutely not desirable, then this change must be accompanied by a major version bump (e.g., releasing as 1.0.0 if coming from 0.x). This clearly signals to users that breaking changes are included, allowing package managers to handle upgrades appropriately.
Example of the desired deprecation approach:

```python
import warnings

class AIOKafkaProducer:
    def __init__(self, *args, api_version=None, **kwargs):
        if api_version is not None:
            warnings.warn(
                "The `api_version` parameter is deprecated and will be removed in a future version. "
                "Please set it via `AIOKafkaConfig(api_version=...)` passed to the `config` parameter.",
                DeprecationWarning,
                stacklevel=2
            )
            # Internal logic to handle the config object...
        # ... rest of init
```
This approach balances the library's evolution with the stability needs of its users. Thank you for your consideration and for maintaining this valuable library.

ods:
### Changes

Fixes #1147

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Changelog
 Bugfixes:
 
 * Fix type annotation for `AIOKafkaAdminClient` (issue #1148)
+* Return back and deprecate `api_version` parameter in client classes
+  (issue #1147)
 
 
 0.13.0 (2026-01-02)
--- a/aiokafka/admin/client.py
+++ b/aiokafka/admin/client.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import warnings
 from collections import defaultdict
 from ssl import SSLContext
 from typing import Any
@@ -92,6 +93,7 @@ class AIOKafkaAdminClient:
         metadata_max_age_ms: int = 300000,
         security_protocol: str = "PLAINTEXT",
         ssl_context: SSLContext | None = None,
+        api_version: None = None,
         sasl_mechanism: str = "PLAIN",
         sasl_plain_username: str | None = None,
         sasl_plain_password: str | None = None,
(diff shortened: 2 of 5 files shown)


ods:
@vmaurin Thanks! Returned it in all client classes. Didn't touch `AIOKafkaConnection`, as I guess nobody uses it directly.

vmaurin:
@ods Looks good to me

PR #1159, opened 2026-04-21
Client rack support

GlebShipilov:
Add support for KIP-392 (fetch from closest replica) so consumers can read from a same-rack follower instead of the partition leader, eliminating cross-AZ / cross-region traffic when the cluster is configured with a rack-aware replica.selector.class.

- AIOKafkaConsumer: new client_rack kwarg (default "" keeps current behaviour). When set, the rack id is sent in FetchRequest v11+ and the broker may respond with a preferred read replica.
- Protocol: added FetchRequest/FetchResponse versions 5–11. v11 carries the rack_id field on requests and preferred_read_replica on responses; intermediate versions add LSO, leader epoch, and incremental fetch session fields needed to get there.
- Fetcher: new per-partition cache tp -> (preferred_replica_node_id, expires_at) with TTL bound to metadata_max_age_ms. _select_read_replica honours the cache when valid, otherwise falls back to the leader. Cache is invalidated on NotLeaderForPartition, UnknownTopicOrPartition, OffsetOutOfRange, or when the chosen broker disappears from cluster metadata.
- Correct handling of preferred_read_replica == -1: the meaning depends on the responder (per KIP-392). -1 from the leader means "no preferred replica, read from me" → drop the cache. -1 from the previously selected follower means "I am still the right one, keep using me" → keep the cache. Without this distinction the consumer oscillated between leader and follower on every fetch and rack-aware routing produced no measurable benefit.
- Added documentation and unit tests covering the cache TTL, both -1 cases, error-driven invalidation, fall-through to the leader when no live preferred replica is known, and back-compat with brokers that only speak < v11.

Verified against a 3-rack test Kafka cluster: 100% of records were served from same-rack brokers, with the expected mix of leader and follower fetches per partition.

Checklist

- [x]  I think the code is well written
- [x]  Unit tests for the changes exist
- [x]  Documentation reflects the changes
- [x]  Add a new news fragment into the CHANGES folder (1159.feature)


ods:
@vmaurin, could you please take a look too?

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,292 @@
...
+FetchRequest to the correct broker after receiving a preferred_read_replica
+hint from the leader?"
+"""
+
+from __future__ import annotations
+
+import time
+from unittest import mock
+
+import pytest
```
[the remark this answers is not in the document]
GlebShipilov: fixed

aiokafka/consumer/consumer.py
```diff
@@ -328,6 +335,8 @@ def __init__(
         self._isolation_level = isolation_level
         self._rebalance_timeout_ms = rebalance_timeout_ms
         self._max_poll_interval_ms = max_poll_interval_ms
+        self._client_rack = client_rack or ""
```
ods:
```suggestion
        self._client_rack = client_rack
```

aiokafka/consumer/fetcher.py
```diff
@@ -408,6 +415,11 @@ def __init__(
         self._retry_backoff = retry_backoff_ms / 1000
         self._subscriptions = subscriptions
         self._default_reset_strategy = OffsetResetStrategy.from_str(auto_offset_reset)
+        self._client_rack = client_rack or ""
```
ods:
```suggestion
        self._client_rack = client_rack
```

aiokafka/protocol/fetch.py
```diff
@@ -541,12 +560,63 @@ def build(
                 self._topics,
             )
 
+        if api_version >= 5:
+            # v5+ adds per-partition `log_start_offset`. v9+ also adds
+            # `current_leader_epoch`. v7+ adds incremental fetch session
+            # fields and forgotten_topics_data. v11 adds top-level rack_id.
+            include_leader_epoch = api_version >= 9
+            partitions_by_topic: list[tuple[str, list[tuple[int, ...]]]] = []
+            for topic, parts in self._topics:
```
ods: _nit:_ `parts` -> `partitions`

aiokafka/protocol/fetch.py
```diff
@@ -541,12 +560,63 @@ def build(
...
+                                partition,
+                                offset,  # fetch_offset
+                                -1,  # log_start_offset (consumer)
+                                max_bytes,
+                            )
+                        )
+                partitions_by_topic.append((topic, new_parts))
+
+            args: list = [
+                -1,  # replica_id
```
vmaurin, line 590: I am a bit unsure if we should still send -1 when consuming from a follower, but it seems to be valid when I look to the java client code https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractFetch.java#L317

I am just a bit confused by this sentence in the KIP

> the FetchRequest schema has field for the replica id. Consumers typically use the sentinel -1, which indicates that fetching is only allowed from the leader. A lesser known sentinel is -2, which was originally intended to be used for debugging and allows fetching from followers. We propose to let the consumer use this to indicate the intent to allow fetching from a follower.

aiokafka/protocol/fetch.py
```diff
@@ -517,21 +519,38 @@ def __init__(
...
         self._isolation_level = isolation_level
         self._topics = topics
+        self._rack_id = rack_id
 
     @property
     def topics(self) -> list[tuple[str, list[tuple[int, int, int]]]]:
         return self._topics
 
+    @property
+    def rack_id(self) -> str:
```
vmaurin: nit: I am not sure it is used currently

aiokafka/protocol/fetch.py
```diff
@@ -541,12 +560,63 @@ def build(
...
+                    ]
+                )
+            args.append(partitions_by_topic)
+            if api_version >= 7:
+                args.append([])  # forgotten_topics_data
+            if api_version >= 11:
+                args.append(self._rack_id)
+            return request_struct_class(*args)
+
         if self._isolation_level:
```
vmaurin, line 610: Should we raise a similar exception is someone is setting a rack_id, while the broker doesn't support v11 ?
* yes, you see the error and you can maybe fix your consumer config / update the broker. On the other side, your consumer keeps crashing
* no, the consumer "works", but it will never consume from a follower in the same rack

aiokafka/protocol/fetch.py
```diff
@@ -541,12 +560,63 @@ def build(
...
+            include_leader_epoch = api_version >= 9
+            partitions_by_topic: list[tuple[str, list[tuple[int, ...]]]] = []
+            for topic, parts in self._topics:
+                new_parts = []
+                for partition, offset, max_bytes in parts:
+                    if include_leader_epoch:
+                        new_parts.append(
+                            (
+                                partition,
+                                -1,  # current_leader_epoch (unknown)
```
ods, line 572: (off-topic) It's a pity we don't follow KIP-320: this may lead to silently skipped data

aiokafka/protocol/fetch.py
```diff
@@ -517,21 +519,38 @@ def __init__(
         max_bytes: int,
         isolation_level: int,
         topics: list[tuple[str, list[tuple[int, int, int]]]],
+        rack_id: str = "",
```
vmaurin, line 522: I do believe the default (and also in the consumer) should be None and not an empty string. In the protocol, a null string is encoded as -1, that is different from a empty string that is encoded as a 0. I am not sure it makes a difference for the broker, but it might.

In the Java client, it looks like the null string will be passed when the consumer rack id is not specified (so -1)

aiokafka/consumer/fetcher.py
```diff
@@ -657,6 +670,59 @@ def _get_actions_per_node(self, assignment):
...
+            node_id, expires_at = cached
+            if time.monotonic() < expires_at and (
+                self._client.cluster.broker_metadata(node_id) is not None
+            ):
+                return node_id
+            # Expired or node disappeared from metadata
+            self._preferred_read_replica.pop(tp, None)
+        return self._client.cluster.leader_for_partition(tp)
+
+    def _update_preferred_read_replica(self, tp, node_id, *, responder_node_id=None):
```
vmaurin: nit: node_id could be named "preferred_read_replica" to make the code a bit more readable

aiokafka/consumer/fetcher.py
```diff
@@ -657,6 +670,59 @@ def _get_actions_per_node(self, assignment):
...
+          answered with ``-1``, it is telling us "I am still the right one,
+          keep using me." Dropping the cache here would bounce the consumer
+          back to the leader on every fetch, defeating the purpose of KIP-392.
+
+        ``responder_node_id`` identifies the broker that produced the response.
+        When unknown, we fall back to the previous (more conservative) behavior
+        of dropping the cache on ``-1``.
+        """
+        if node_id is None or node_id == -1:
+            if responder_node_id is None:
```
vmaurin: I am not sure that is a use case. When we call this method, we just sent a FetchRequest to a node, so we should know which one

aiokafka/consumer/fetcher.py
```diff
@@ -408,6 +415,11 @@ def __init__(
         self._retry_backoff = retry_backoff_ms / 1000
         self._subscriptions = subscriptions
         self._default_reset_strategy = OffsetResetStrategy.from_str(auto_offset_reset)
+        self._client_rack = client_rack or ""
+        # KIP-392: cache of preferred read replica per partition.
+        # tp -> (node_id, expires_at_monotonic)
+        self._preferred_read_replica: dict[TopicPartition, tuple[int, float]] = {}
+        self._preferred_replica_ttl = metadata_max_age_ms / 1000
```
vmaurin: nit: metadata_max_age_ms is already passed to the client, that fetch metadata and expose it part of the "ClusterMetadata" object. Maybe it should be read from here (with an helper ?). It sounds like this in the java client

vmaurin (commented):
* Implementation of the KIP-392 sounds correct (added few comments)
* v11 fetch request/response is not using flexible version yet, so we should be fine here
* I am not so sure about the implication of KIP-320 (v9) and incremental fetch support (v7), but it sounds we pass default value there, so maybe we should just be sure tests are working well with brokers using these versions of the protocol (probably already in the list)

aiokafka/protocol/fetch.py
GlebShipilov: yes, the Java client does the same. -1 means "I'm a consumer" not "only fetch from leader".

aiokafka/protocol/fetch.py
GlebShipilov: No, silent fallback is what Java does. But worth a log warning.

aiokafka/protocol/fetch.py
ods: In the protocol `rack_id` field in `Fetch` request is defined as `STRING`, not `NULLABLE_STRING`

GlebShipilov: Will change to None

aiokafka/consumer/fetcher.py
GlebShipilov: Sure, will do

aiokafka/consumer/fetcher.py
GlebShipilov: I will remove the None fallback, make it required

aiokafka/consumer/fetcher.py
GlebShipilov: Agree

aiokafka/protocol/fetch.py
vmaurin: @ods you are right, I have overlooked the definition
```
{ "name": "RackId", "type":  "string", "versions": "11+", "default": "", "ignorable": true,
      "about": "Rack ID of the consumer making this request."}
```

So it should probably be left empty string (at consumer def level, or when we actually do send the request)

aiokafka/consumer/fetcher.py
```diff
@@ -657,12 +683,82 @@ def _get_actions_per_node(self, assignment):
...
+        ``responder_node_id`` identifies the broker that produced the response.
+        """
+        if preferred_read_replica is None or preferred_read_replica == -1:
+            leader = self._client.cluster.leader_for_partition(tp)
+            # Only drop the cache if the leader itself told us there is no
+            # preferred replica. A `-1` from the previously chosen follower
+            # means "I am still the right choice" — keep it.
+            if responder_node_id == leader:
+                self._preferred_read_replica.pop(tp, None)
+            return
```
vmaurin: Maybe this whole block is not needed:
* either we just sent the fetch request to the leader meaning that the cache was empty, or pointing to the leader, and basically the leader just tell us "keep going"
* or we just sent the request to a replica, that return -1 to say "don't change the cache"

So maybe the whole function could just update the cache when 

```
if preferred_read_replica is not None and preferred_read_replica >= 0:
    self._preferred_read_replica[tp] = (
        preferred_read_replica,
        time.monotonic() + self._preferred_replica_ttl,
    )
```

The cache invalidation is happening on error or when the timer is expiring, or when a broker indicate us to look in other place

GlebShipilov: Yes, you are right. The two cases you outlined are exhaustive (the routing layer reads the same cache, so responder == leader always implies the cache is already empty or already pointing at the leader), and TTL + transport-failure eviction + the routing-error path already cover real invalidation. 
I'll collapse the function to the update-only branch and update test_minus_one_from_leader_clears_cache to assert "no entry is created" instead of "entry is dropped". Pushing in the next commit.

ods: @GlebShipilov Please also add this justification in a comment

GlebShipilov:
I can see that there are some failed tests.

All 5 failing CI jobs **are not caused by the rack-awareness changes in this PR.** 
They fall into two categories:

1. Producer timeout flake (Python 3.11 / Kafka 2.8.1)
test_producer_send_leader_notfound fails because its request_timeout_ms=200 is too tight. The broker doesn't finish auto-creating the topic within 200 ms on a busy GHA runner. The two accompanying teardown ERRORs (test_producer_warn_unclosed) are cascading resource warnings from the un-closed producer. A re-run would likely pass. 
I can bump request_timeout_ms in that test, or just leave it as-is, maybe re-run will help.
Also, what I can do to fix it: call await self.wait_topic(producer.client, self.topic) before the mock, so metadata is already cached when send() runs. The 200ms timeout then only applies to the actual send (which is what the test intends to exercisem, the leader-not-found retry path), not the topic creation. 


2. Python 3.14 transactional consumer regression (all four 3.14 jobs)
test_consumer_transactional_commit and test_consumer_transactional_abort fail with AssertionError: 0 != 3 on every Kafka version in the matrix (0.11.0.3, 1.1.1, 2.1.1, 2.2.2) but only on Python 3.14. The same tests pass on Python 3.11. The assertions at lines 74-75 and 133-134 expect last_stable_offset / highwater to be 3 after a transaction commit/abort, but get 0, suggesting the consumer's fetch loop doesn't complete an offset update in time under Python 3.14's changed asyncio scheduler. 
This looks like a pre-existing asyncio compat problem. I can increase the asyncio.sleep(1) in those tests, add a retry/poll loop for the offset assertions, or mark them xfail on 3.14, but I'd rather not mask it if you'd prefer a proper upstream investigation.

What do you think?

aiokafka/consumer/fetcher.py
```diff
@@ -657,12 +683,75 @@ def _get_actions_per_node(self, assignment):
...
+          cached follower), so there is nothing to invalidate.
+        * If we sent the request to the cached **follower**, ``-1`` means
+          "stay with me" — preserving the cache is the desired behaviour.
+
+        Real cache invalidation happens elsewhere: TTL expiry in
+        :meth:`_select_read_replica`, transport-failure eviction in
+        :meth:`_invalidate_preferred_read_replica_for_node`, and routing
+        errors (e.g. ``NotLeaderForPartition``) in the fetch error path.
+        """
+        if preferred_read_replica is not None and preferred_read_replica >= 0:
```
ods: Is it ever possible to be `None`?

GlebShipilov: actually, it's a leftover from some previous version, in current code it can never be None

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,524 @@
+"""Tests for rack-aware replica selection (KIP-392).
+
+Each test verifies an observable behaviour of the feature, not internal
+plumbing. The core question answered: "does the consumer route its next
+FetchRequest to the correct broker after receiving a preferred_read_replica
+hint from the leader?"
+"""
+
+from __future__ import annotations
+
```
ods: _nit:_ Looks like it's not needed here
```suggestion
```

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,524 @@
...
+    """
+    client = mock.Mock()
+    client._loop = mock.Mock()
+    client.cluster = mock.Mock()
+    client._metadata_max_age_ms = metadata_max_age_ms
+    subscriptions = mock.Mock()
+    subscriptions.register_fetch_waiters = mock.Mock()
+
+    with (
+        mock.patch("aiokafka.consumer.fetcher.create_task"),
```
ods: This is fragile and I believe we don't need to patch it, if we use empty async function as `_fetch_requests_routine` replacement. But then we'd have to make all tests using `_make_fetcher` async.

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,524 @@
...
+        # Second call should NOT log again.
+        caplog.clear()
+        with caplog.at_level(logging.WARNING):
+            await fetcher._proc_fetch_request(assignment, 1, request)
+        assert not any("client_rack" in msg for msg in caplog.messages)
+
+    @pytest.mark.asyncio
+    async def test_no_warning_when_rack_not_set(self, caplog):
+        """If client_rack is not set, no warning even when broker < v11."""
+        fetcher = _make_fetcher(client_rack=None)
```
ods: It's declared as `str` here and in `Fetcher`. Did you mean an empty string?

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,524 @@
...
+            max_wait_time=100,
+            min_bytes=1,
+            max_bytes=1024,
+            isolation_level=0,
+            topics=[("t", [(0, 42, 1024)])],
+            rack_id="us-east-1a",
+        )
+        built = req.build(FetchRequest_v9)
+        obj = built.to_object()
+        assert "rack_id" not in obj
```
ods: There is no such requirement

```suggestion
```

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,524 @@
...
+# ---------------------------------------------------------------------------
+# 5. Warning when client_rack is set but broker < v11
+# ---------------------------------------------------------------------------
+
+
+class TestRackVersionWarning:
+    """A one-shot warning must be logged when client_rack is set but the
+    broker only supports FetchRequest < v11."""
+
+    @pytest.mark.asyncio
```
ods: _nit:_ Here and in other places: no need in marker, we're using "auto" mode.

```suggestion
```

tests/test_rack_awareness.py
```diff
@@ -0,0 +1,524 @@
...
+        fetcher._update_preferred_read_replica(tp, 7)
+        assert tp in fetcher._preferred_read_replica
+
+        # Simulate a transport failure when sending to follower 7.
+        fetcher._client.send = mock.AsyncMock(
+            side_effect=Errors.KafkaConnectionError("boom")
+        )
+
+        # Sleep is awaited inside the error path; make it instant.
+        with mock.patch("aiokafka.consumer.fetcher.asyncio.sleep", mock.AsyncMock()):
```
ods:
```suggestion
        with mock.patch.object(asyncio, "sleep", mock.AsyncMock()):
```

docs/consumer.rst
```diff
@@ -546,6 +546,61 @@ those messages would not be returned by the consumer and yet would have valid
...
+   (`KIP-36`_) and advertises the rack in metadata responses.
+2. Each broker must have a replica selector enabled, e.g.
+   ``replica.selector.class=org.apache.kafka.common.replica.RackAwareReplicaSelector``.
+   Without this, the broker always returns ``preferred_read_replica = -1``
+   ("keep fetching from the leader") regardless of the rack id sent by the
+   consumer.
+
+Some caveats to keep in mind:
+
+* The "preferred" replica is only used **while it stays in the ISR**. If the
```
ods: What is "ISR"?

vmaurin: Kafka acronym for "In Sync Replicas". So group of followers that are considered in sync with the leader

ods:
Thanks, @GlebShipilov — really high-quality work! It looks ready to merge to me. @vmaurin, your opinion?

vmaurin:
> Thanks, @GlebShipilov — really high-quality work! It looks ready to merge to me. @vmaurin, your opinion?

Yes, it looks good, thanks @GlebShipilov ! I haven't got a chance to test it, but I feel it could be merged to the main branch.

I think the most important thing before a release would be to check if it doesn't break for people not using the rack awareness feature

ods:
> I think the most important thing before a release would be to check if it doesn't break for people not using the rack awareness feature

We already did this for our setup. And I'm going to release beta version first, so that more people can do this.

PR #1162, opened 2026-05-19
Revisit tagged fields

vmaurin:
Instead of a dedicated field type, treat tagged fields as an option of schema, to then be able to specify them with a name (and their tag) so it is easier to maniuplate on the Struct class.

It tries to reproduce the generated code of the official Java client with the `TaggedFieldsSection` class.

It also drops the FLEXIBLE_VERSION boolean detected by the presence of tagged_fields in the request schema

### Checklist

### DIFF
--- a/aiokafka/protocol/admin.py
+++ b/aiokafka/protocol/admin.py
@@ -17,7 +17,6 @@ from .types import (
     Int64,
     Schema,
     String,
-    TaggedFields,
 )
 
 
@@ -1467,18 +1466,17 @@ class AlterPartitionReassignmentsResponse_v0(Response):
                         ("partition_index", Int32),
                         ("error_code", Int16),
                         ("error_message", CompactString("utf-8")),
-                        ("tags", TaggedFields),
+                        tagged_fields=(),
                     ),
                 ),
-                ("tags", TaggedFields),
+                tagged_fields=(),
             ),
         ),
-        ("tags", TaggedFields),
+        tagged_fields=(),
     )
 
 
 class AlterPartitionReassignmentsRequest_v0(RequestStruct):
-    FLEXIBLE_VERSION = True
     API_KEY = 45
     API_VERSION = 0
     RESPONSE_TYPE = AlterPartitionReassignmentsResponse_v0
@@ -1493,13 +1491,13 @@ class AlterPartitionReassignmentsRequest_v0(RequestStruct):
                     CompactArray(
                         ("partition_index", Int32),
                         ("replicas", CompactArray(Int32)),
-                        ("tags", TaggedFields),
+                        tagged_fields=(),
                     ),
                 ),
-                ("tags", TaggedFields),
+                tagged_fields=(),
             ),
         ),
-        ("tags", TaggedFields),
+        tagged_fields=(),
     )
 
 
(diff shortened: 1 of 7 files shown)


vmaurin:
@GlebShipilov @ods Ready for review

aiokafka/protocol/admin.py
```diff
@@ -17,7 +17,6 @@
     Int64,
     Schema,
     String,
-    TaggedFields,
```
GlebShipilov, line 20: TaggedFields removed without deprecation shim. from aiokafka.protocol.types import TaggedFields was public API. Downstream code doing this will now get ImportError. 
I believe, it would be better to add a module-level __getattr__ (or keep the class as a deprecated thin wrapper) that emits DeprecationWarning on access. Low effort, prevents silent breakage for users upgrading from 0.14.

aiokafka/protocol/types.py
```diff
@@ -185,22 +185,114 @@ def decode(cls, data: BytesIO) -> bool:
...
+                else:
+                    result.append(None)
+
+        return tuple(result)
+
+    @staticmethod
+    def _encode_tagged_fields(value: list[tuple[int, bytes]]) -> bytes:
+        ret = UnsignedVarInt32.encode(len(value))
+        for k, v in value:
+            assert isinstance(k, int) and k >= 0, f"Key {k} is not a positive integer"
```
GlebShipilov: assert used for runtime validation in _encode_tagged_fields:
assert isinstance(k, int) and k >= 0, f"Key {k} is not a positive integer"
assert is stripped under python -O. Since tag ordering is already validated in __init__ this is probably dead code in practice, but if you want the safety net, make it if not ...: raise ValueError(...).

aiokafka/protocol/types.py
```diff
@@ -185,22 +185,114 @@ def decode(cls, data: BytesIO) -> bool:
...
+            ret += UnsignedVarInt32.encode(k)
+            ret += UnsignedVarInt32.encode(len(v))
+            ret += v
+        return ret
+
+    @staticmethod
+    def _decode_tagged_fields(data: BytesIO) -> dict[int, bytes]:
```
GlebShipilov, line 284: _decode_tagged_fields doesn't validate tag ordering on the wire. The old code raised on out-of-order tags. The new code silently accepts them into a dict. Being lenient on decode is defensible for forward-compat, but worth a comment explaining the intent. Otherwise it looks like an accidental omission.

GlebShipilov (commented):
This is a great improvement over the previous PR!
There are just a few things, which I'd like to address

aiokafka/protocol/admin.py
ods: I believe we can safely assume the whole `protocol` package not a public API. In any case, I see no `TaggedFields` usage outside `aiokafka` itself throughout all public repositories on GitHub.

aiokafka/protocol/types.py
vmaurin: Nice catch, I have removed the unnecessary check

aiokafka/protocol/types.py
vmaurin: It was voluntary, so I added a comment to the code

ods (approved):

PR #1163, opened 2026-05-24
Fix getone/getmany raising AttributeError when consumer not started

### LINKED ISSUE #1102: If consumer isn't started, `getmany()` fails with `AttributeError`
**Describe the bug**
If a consumer is created and the `genmany()` method is called, an AttributeError results.

**Expected behaviour**
A user friendly error that explains the problem should be thrown.

 - aiokafka version: 0.12.0
 - Kafka Broker version: 3.9.0
 - Other information (Confluent Cloud version, etc.): n/a

**Reproducible example**
```python
async def err_on_genmany():
    kc = AIOKafkaConsumer(
        "topichere",
        bootstrap_servers=f"localhost:{kcon.port}",
        auto_offset_reset="earliest"
    )
    await kc.getmany()
    kc.stop()


if __name__ == "__main__":
    asyncio.run(err_on_genmany())
```

Results in the output:
```
  File "/home/<user>test_manual/kafka_notifier_manual_test.py", line 130, in <module>
    asyncio.run(err_on_genmany())
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/<user>test_manual/kafka_notifier_manual_test.py", line 121, in err_on_genmany
    await kc.getmany()
  File "/home/<user>/.local/share/virtualenvs/<dir>-2sgYN4mf/lib/python3.11/site-packages/aiokafka/consumer/consumer.py", line 1208, in getmany
    self._coordinator.check_errors()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'check_errors'
```

vincere-mori:
## Problem

Calling `getone()` or `getmany()` on an `AIOKafkaConsumer` that has been created but not yet started raises an opaque `AttributeError`:

```
AttributeError: 'NoneType' object has no attribute 'check_errors'
```

Repro:
```python
consumer = AIOKafkaConsumer("topic", bootstrap_servers="localhost:9092")
await consumer.getmany()  # crashes with AttributeError
```

The problem: `self._coordinator` is `None` until `start()` is called, but the only guard in `getone()`/`getmany()` checks `self._closed`, which is `False` from `__init__`. So the code reaches `self._coordinator.check_errors()` on a `None` object.

## Fix

Add an explicit `if self._fetcher is None` check in both `getone()` and `getmany()`. `self._fetcher` is the first attribute assigned in `start()`, making it a reliable "started" indicator. When the consumer hasn't been started, raise `IllegalStateError` with an actionable message directing the user to call `await consumer.start()` or use the `async with` context manager.

Closes #1102

### DIFF
--- a/aiokafka/consumer/consumer.py
+++ b/aiokafka/consumer/consumer.py
@@ -1136,6 +1136,11 @@ class AIOKafkaConsumer:
         assert all(isinstance(k, TopicPartition) for k in partitions)
         if self._closed:
             raise ConsumerStoppedError()
+        if self._fetcher is None:
+            raise IllegalStateError(
+                "AIOKafkaConsumer is not started. Use `await consumer.start()` "
+                "or the `async with` context manager."
+            )
 
         # Raise coordination errors if any
         self._coordinator.check_errors()
@@ -1183,6 +1188,11 @@ class AIOKafkaConsumer:
         assert all(isinstance(k, TopicPartition) for k in partitions)
         if self._closed:
             raise ConsumerStoppedError()
+        if self._fetcher is None:
+            raise IllegalStateError(
+                "AIOKafkaConsumer is not started. Use `await consumer.start()` "
+                "or the `async with` context manager."
+            )
 
         if max_records is not None and (
             not isinstance(max_records, int) or max_records < 1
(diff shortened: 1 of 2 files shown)


vmaurin:
LGTM ! You should document your change in https://github.com/aio-libs/aiokafka/blob/master/CHANGES.rst (and maybe in the docs ?)

PR #1164, opened 2026-05-25
fix: add retry backoff for NotLeaderForPartitionError in _proc_fetch_request

### LINKED ISSUE #1155: Missing retry backoff in Fetcher for NotLeaderForPartitionError causes tight fetch loop during broker restarts
When `Fetcher._proc_fetch_request` receives [NotLeaderForPartitionError or UnknownTopicOrPartitionError](https://github.com/aio-libs/aiokafka/blob/v0.13.0/aiokafka/consumer/fetcher.py#L765) in the per-partition response loop, it calls `self._client.force_metadata_update()` but does not apply any backoff/sleep before the next fetch cycle.

Normally Kafka broker waits up to fetch.max.wait.ms (default 500ms) before responding to a fetch request. However, when a partition has no leader (e.g. during broker restart or partition reassignment), the broker returns `NotLeaderForPartitionError` immediately without entering the wait phase. The response time drops to just the network round-trip (~10-20ms).

Without a backoff on the client side, the fetch loop spins as fast as the network allows, effectively DDoS-ing the Kafka cluster with fetch requests. This causes memory usage to spike in the consumer process.

Note that the general KafkaError exception handler at the top of the method (added in #534) does include `await asyncio.sleep(self._retry_backoff)`, but the per-partition error handling branch for NotLeaderForPartitionError inside the response processing loop does not.

### Expected behaviour
After receiving NotLeaderForPartitionError or UnknownTopicOrPartitionError, the fetcher should apply `self._retry_backoff` sleep (same as it does for general KafkaError), giving the cluster time to elect new partition lleaders without being overwhelmed by requests.

```python
  elif error_type in (
      Errors.NotLeaderForPartitionError,
      Errors.UnknownTopicOrPartitionError,
  ):
      self._client.force_metadata_update()
      await asyncio.sleep(self._retry_backoff)  # missing backoff
```

vincere-mori:
## Problem

Closes #1155.

When a broker returns `NotLeaderForPartitionError` or `UnknownTopicOrPartitionError` for one partition in a fetch response, aiokafka forces a metadata update but immediately retries that partition. During leader changes this can create a tight fetch loop while metadata is still propagating.

## Fix

This PR adds a per-partition fetch backoff after leader metadata errors:

- `TopicPartitionState` tracks a fetch-backoff deadline
- `_proc_fetch_request` requests `_retry_backoff` only for partitions that returned stale-leader metadata errors
- `_get_actions_per_node` skips only those backed-off partitions while still allowing other partitions to fetch

This avoids blocking unrelated partitions and still prevents repeated immediate fetches for partitions waiting on fresh metadata.

## Tests

- `AIOKAFKA_NO_EXTENSIONS=1 PYTHONPATH=. pytest tests/test_fetcher.py -q`
- `AIOKAFKA_NO_EXTENSIONS=1 PYTHONPATH=. pytest tests/test_rack_awareness.py -q`
- `ruff check aiokafka/consumer/fetcher.py aiokafka/consumer/subscription_state.py tests/test_fetcher.py tests/test_rack_awareness.py`

### DIFF
--- a/aiokafka/consumer/fetcher.py
+++ b/aiokafka/consumer/fetcher.py
@@ -590,6 +590,7 @@ class Fetcher:
         fetchable = collections.defaultdict(list)
         awaiting_reset = collections.defaultdict(list)
         backoff_by_nodes = collections.defaultdict(list)
+        fetch_backoffs = []
         resume_futures = []
         invalid_metadata = False
 
@@ -634,6 +635,10 @@ class Fetcher:
             elif tp_state.paused:
                 resume_futures.append(tp_state.resume_fut)
             else:
+                backoff = tp_state.fetch_backoff()
+                if backoff:
+                    fetch_backoffs.append(backoff)
+                    continue
                 position = tp_state.position
                 fetchable[node_id].append((tp, position))
                 log.debug(
@@ -669,10 +674,10 @@ class Fetcher:
             )
             fetch_requests.append((node_id, req))
 
-        if backoff_by_nodes:
+        if backoff_by_nodes or fetch_backoffs:
             # Return min time till any node will be ready to send event
             # (max of it's backoffs)
-            backoff = min(map(max, backoff_by_nodes.values()))
+            backoff = min([*map(max, backoff_by_nodes.values()), *fetch_backoffs])
         else:
             backoff = self._fetcher_timeout
         return (
@@ -888,6 +893,7 @@ class Fetcher:
                         Errors.UnknownTopicOrPartitionError,
                     ):
                         self._client.force_metadata_update()
+                        tp_state.request_fetch_backoff(self._retry_backoff)
                     elif error_type is Errors.OffsetOutOfRangeError:
                         if self._default_reset_strategy != OffsetResetStrategy.NONE:
                             tp_state.await_reset(self._default_reset_strategy)
--- a/aiokafka/consumer/subscription_state.py
+++ b/aiokafka/consumer/subscription_state.py
@@ -474,6 +474,7 @@ class TopicPartitionState:
 
         self._paused = False
         self._resume_fut = None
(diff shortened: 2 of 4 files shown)


aiokafka/consumer/fetcher.py
```diff
@@ -743,6 +743,7 @@ def _invalidate_preferred_read_replica_for_node(self, node_id, request):
 
     async def _proc_fetch_request(self, assignment, node_id, request):
         needs_wakeup = False
+        notleader_seen = False
```
vmaurin: nit: the boolean name is a bit weird, maybe it could be "partitions_with_errors" or something similar and also be moved just before the for loop

vmaurin (commented):
I am not sure on the overall approach here:
* to reach this point, we did successfully send a fetch request to a broker
* the broker is indicated us that one or many of the queried partitioned are not available on its side

So it is a bit unnecessary to sleep on the whole fetch routine especially if there are still partition valid, from which we can grep data. Ideally, on such case, we should try instead to dissociate this partition from this specific brokers, eventually with a retry backoff  (i.e "don't try again on this partition for 100ms")

Maybe something to mark on TopicPartitionState that could be used when preparing the fetch request

aiokafka/consumer/subscription_state.py
```diff
@@ -554,6 +555,12 @@ def seek(self, position: int):
     def wait_for_position(self):
         return shield(self._position_fut)
 
+    def request_fetch_backoff(self, backoff: float):
+        self._fetch_backoff_until = time.monotonic() + backoff
+
+    def fetch_backoff(self):
```
vmaurin, line 561: nit: could be declared as a property like "paused"

vmaurin (commented):
The new approach looks good to me, maybe @vanchaxy could give it a try ?

PR #1169, opened 2026-07-04
Make key-based partition selection deterministic

### LINKED ISSUE #1127: ‎AIOKafkaProducer.send‎ shuffles partitions when running in PyPy environments
**Describe the bug**

This bug is not specific to PyPy, but because of the way CPython handles `set` instances, it will only be reproducible in CPython with a large number of partitions. In PyPy (RPython) this bug can be easily reproduced with a few partitions, such as 20 in my case.

It is hard to provide a small example for this issue, because it requires a Kafka server, but the issue itself is not hard to reproduce and happens all the time.

Kafka provides `aiokafka` with a list of partitions in arbitrary order and when `aiokafka` computes a partition from the supplied key in `provider.send()`, it relies on the specific ordering of a `set` instance, which is a hash table and has no specific order. Here's the chain of calls where this occurs.

`producer.send(topic, value, key))` calls `AIOKafkaProducer._partition`, which obtains all partitions as follows:
```python
all_partitions = list(self._metadata.partitions_for_topic(topic))
```
`‎ClusterMetadata.partitions_for_topic`, in turn, returns all partitions in a `set`, like this:
```python
return set(self._partitions[topic].keys())
```
The partition then is figured out via the call below, where the partitioner uses the modulo-divided Murmur2 hash value as an index into `all_partitions`, which has indetermined order at this point because this list was constructed from a `set`.
```python
return self._partitioner(serialized_key, all_partitions, available)
```
The reason people running this code in CPython don't see the bug is because of how CPython hashes numbers, which makes it look like a `set` is ordered, but this is just an illusion and it only works for small numbers.

This results in the same keys associated with different partitions, which may happen right away or days later, depending on how Kafka supplies partitions to `aiokafka` and how many rebalancing events takes place. This corrupts continuity of the state attributed to those keys.

**Expected behaviour**

There are two issues here:

* The same key must always end up in the same partition.
* The computed partition obtained outside of the `provider.send()` call must be the same as the one computed for the same key inside `provider.send()`, given that the same partitioner is being used in both cases. 

 - aiokafka version (`python -c "import aiokafka; print(aiokafka.__version__)"`): `'0.12.0'`
 - Kafka Broker version (`kafka-topics.sh --version`): `4.0.0`
 - Other information (Confluent Cloud version, etc.): `aiokafka` is running on Ubuntu 22.04, Python 3.10, PyPy

**Reproducible example**
It is hard to set up an example for this issue because it requires Kafka and a few consumers. However, the fact that `set` instances have indetermined order is well documented and this bug is clearly visible in the code even without a reproducible case.

You can see how CPython sets are different from PyPy sets if you run this code in PyPy and in CPython in the link provided below.

```python
import random

l1: list[int] = [i for i in range(20)]
# simulates how Kafka provides partitions
random.shuffle(l1)

# aiokafka performs this to obtiain all_partitions
s: set[int] = set[int](l1)
l2 = list[int](s)

# will be shuffled in PyPy and will seem sorted in CPython
print(l2)
```

https://ide.judge0.com/

Choose `Python 3.10 (PyPy 7.3.12)` for PyPy and `Python 3.11.2` for CPython. There's no Python 3.10 at this point, which is what I'm running, but Python 3.11 shows the same behavior. You will see that CPython yields a sorted list of partitions, despite the shuffling, and PyPy produces a randomized list, which is what's happening in `aiokafka` running on PyPy.

Worth noting that CPython's ordering is not the intended behavior, but rather a side effect of how CPython "hashes" small numbers., and will not hold for larger partition sets.

Any time a `set` is converted into a list, it must be sorted, which may be expensive if done each time, but in general `all_partitions` just needs to be constructed using a sorted list, which will fix this bug.

arpitjain099:
### Changes

Fixes #1127

`_partition` builds the partition list with `list()` over `partitions_for_topic()`, which returns a set, so the ordering is not stable across processes or interpreters. The default partitioner then uses the murmur2 hash modulo length as an index into that list, so the same key can land on a different partition and the same-key-same-partition guarantee quietly breaks. As the reporter notes, it shows up reliably on PyPy or with a large number of partitions, where CPython small-int hashing no longer makes the set look sorted. Sorting both lists by partition id keeps the index stable without changing which partition a key maps to for a given topic. I added a unit test that runs `_partition` with the partitions in several orders and checks that a fixed key always picks the same one (it fails before this change and passes after).

### Checklist


aiokafka/producer/producer.py
```diff
@@ -419,8 +419,14 @@ def _partition(
...
-        all_partitions = list(self._metadata.partitions_for_topic(topic))
-        available = list(self._metadata.available_partitions_for_topic(topic))
+        # partitions_for_topic() and available_partitions_for_topic() return
+        # sets, whose iteration order is not stable across processes and
+        # interpreters. The default partitioner indexes into this list with
+        # murmur2(key) % len(partitions), so an unstable order would map the
+        # same key to different partitions and break key ordering (issue #1127).
+        # Sort by partition id to keep the mapping deterministic.
+        all_partitions = sorted(self._metadata.partitions_for_topic(topic))
+        available = sorted(self._metadata.available_partitions_for_topic(topic))
```
vmaurin: It is somehow a breaking change in the contract of the partitioner as before it was getting 2 sets and now it is getting two lists.

I think it is not a big deal as they share similar interfaces, but maybe it should be mentioned in the changelog in case people did implement their custom partitioners assuming they are gettings sets

Then, we could have few alternatives
* add a `ordered_partitions_for_topic` on the cluster class, so we don't need to sort something that is already sorted (or we should sort it only one time, when we refresh the metadata). This would return a list, still the same breaking change in term of contract
* not sorting here, but sorting inside the default partitioner, so we keep the contract intact. The issue though is that we are going to sort over and over the same thing

arpitjain099: Thanks for looking at this.

I think the contract is actually unchanged here. `partitions_for_topic()` does return a set, but the line was already doing `list(...)` on it before this PR, so the partitioner has always been handed a list. `sorted()` gives back a list the same way. The only thing that changes is the ordering, not the type, so a custom partitioner relying on set operations would have been broken well before this change.

Your perf point stands though. We're now sorting on every send where it used to be just a set-to-list walk. If that cost matters, caching the sorted order on the cluster at metadata refresh (your first option) looks like the better shape to me, and `partitions_for_topic()` could keep returning the set so nothing else moves. Happy to do that here or leave it as a follow-up, whichever you'd prefer.

vmaurin: Ah right, sorry for the overlook, it is indeed already a list so no breaking change.

About the performance, both implementation are quite different, so I would tend to favor a PR that would expose an ordered list of partition, that is then used here to pass it to the partitioner. Especially in Cluster, we have a dict, that preserve the insertion order, and it seems the brokers reply the partition in an ordered manner (<= this should be verified though), so the implementation could be quite minimal like

```py
    def ordered_partitions_for_topic(self, topic: str) -> list[int] | None:
        if topic not in self._partitions:
            return None
        return list(self._partitions[topic].keys())
```

arpitjain099: That accessor works for me, and sorting once per refresh beats sorting on every send.

One catch on `list(self._partitions[topic].keys())` though. The dict's insertion order is just the order the broker sent, since `update_metadata` builds it with `for p_error, partition, leader, replicas, isr, *_ in partitions` straight off the MetadataResponse. So we'd be swapping set iteration order for broker reply order. That's stable inside one process, but nothing in the protocol promises the same ordering across refreshes or brokers, and two producers picking different partitions for the same key is the actual failure in #1127.

Easiest way to get what you want is to sort where the dict is built, so the insertion order is deterministic to begin with:

```py
for p_error, partition, leader, replicas, isr, *_ in sorted(partitions, key=lambda p: p[1]):
```

Then `ordered_partitions_for_topic` works exactly as you wrote it, `partitions_for_topic()` keeps returning the set, and the sort happens once per metadata refresh instead of once per send.

Want me to fold that into this PR, or keep this one as the minimal fix and do the cluster change separately?

vmaurin: I would fold this into this PR as then you won't need to sort/transform to list in the `_partition` method anymore.

As you mention, I was not able to find in a KIP or documentation a mention that brokers must sort the partitions in the metadata response, so better to sort it explicitly when consuming them

arpitjain099: Folded in, thanks.

Funny thing I hit while doing it: the docstring on `DefaultPartitioner.__call__` already describes `all_partitions` as a "list of all partitions sorted by partition ID", and `available` as "in no particular order". So sorted was the intended contract all along, the cluster just wasn't honoring it. Makes this less of a change than it first looked.

`update_metadata` now sorts as it builds the dict, so insertion order doesn't depend on what the broker happened to send. `ordered_partitions_for_topic` is your version verbatim. I added `ordered_available_partitions_for_topic` next to it since `_partition` needs both sides and the available one was still a set, otherwise you'd be back to a `list()` call in there. It's free now that insertion order is sorted, same filter with a list instead of a set. Both existing set-returning accessors are untouched.

The test now drives a real `ClusterMetadata` with a `MetadataResponse` whose partitions arrive out of order, instead of the fake I had before, which would have just asserted the ordering into existence once the sort moved. Dropped the sort locally to check it actually fails without it.

vmaurin: Look good to me :+1: Maybe @ods for a second look / approval ?

vmaurin: @ods if you are around, could you take a look to this one ?

aiokafka/cluster.py
```diff
@@ -108,6 +108,20 @@ def partitions_for_topic(self, topic: str) -> set[int] | None:
...
+        Arguments:
+            topic (str): topic to check for partitions
+
+        Returns:
+            list: [partition (int), ...] sorted by partition id
+            None if topic not found.
+        """
+        if topic not in self._partitions:
+            return None
+        return list(self._partitions[topic].keys())
```
ods, line 123: There are 2 places where we rely on the dict to preserver the order. This is not obvious, and without proper notes in comments it will be easy to break this with future changes.

aiokafka/cluster.py
```diff
@@ -234,8 +266,15 @@ def update_metadata(self, metadata):
...
+                # deterministic. Nothing in the protocol requires brokers to
+                # return partitions in a fixed order, and ordered_*_for_topic()
+                # rely on that insertion order to keep key-based partitioning
+                # stable across metadata refreshes and processes (issue #1127).
                 # Starting with v5, MetadataResponse contains more than 5 fields
-                for p_error, partition, leader, replicas, isr, *_ in partitions:
+                for p_error, partition, leader, replicas, isr, *_ in sorted(
+                    partitions, key=lambda p: p[1]
+                ):
                     _new_partitions[topic][partition] = PartitionMetadata(
```
ods, line 278: I see we also have `with_partitions` method. Am I correct it may break the order? 

This method is a legacy of merging python-kafka code and I can't find any use of it (neither here, nor in public GitHub repositories). Probably we can remove it. If so, we have to mark it as a breaking change in changelog.

tests/test_partitioner.py
```diff
@@ -41,3 +44,48 @@ def test_murmur2_not_ascii():
...
+
+class _FakeProducer:
+    # _partition() only touches these two attributes, so we can drive the real
+    # method without spinning up a broker.
+    def __init__(self, cluster):
+        self._metadata = cluster
+        self._partitioner = DefaultPartitioner()
+
+
+def test_partition_selection_is_order_independent():
```
ods, line 71: This test also passes without fix applied, therefore it doesn't reproduce the problem

CHANGES/1127.bugfix
ods, line 1: As I understand, users we'll start getting different assignment during update. May it break something? Do we need to warn about it in changelog?

PR #1171, opened 2026-07-11
Add lightweight producer metrics collector

### LINKED ISSUE #1166: Add a lightweight metrics collector protocol for the producer
## Motivation

When end-to-end produce latency is high, an application-level timer
wrapped around `send_and_wait()` cannot tell whether the cause is network
RTT, replication wait on the broker side, producer-side accumulator
back-pressure, retries, or `linger_ms` batching. Each of these has a
different remediation path (cluster topology, broker tuning, producer
config, application code) and a different operational owner. Without
visibility into the producer's internal stages, every investigation
starts from scratch with guesswork or ad-hoc instrumentation.

The Java client exposes ~25 producer metrics that segment this latency
directly — `record-queue-time-avg/max`, `request-latency-avg/max`,
`bufferpool-wait-time-total`, `record-retry-rate`, and so on — and they
are the first thing operators reach for. aiokafka exposes none, which
forces every team running it in production to write the same custom
timing wrappers and still end up unable to distinguish the same causes.

This proposal aims to close that gap with a minimal, pluggable API — **not**
by re-introducing the heavyweight `Sensor`/`Stat`/`Histogram` framework
that was removed in #1117.

## Prior art in this repo

- #38 (open since 2016) — "Add metrics". Original intent was to reuse the
  python-kafka `Sensor` framework, but it was never completed.
- #562 (open since 2019) — "Recommended way to expose metrics to Prometheus".
- #862 (open since 2022) — "Add compatibility with OpenTelemetry". Blocked
  by the lack of hook points.
- #1117 (merged 2025-07) — removed 1969 lines of vendored kafka-python
  metrics code. This proposal deliberately does **not** bring any of it back.

## Proposal: callback-based `ProducerMetricsCollector` protocol

A single new module `aiokafka/metrics.py` defines a Protocol with sync
callback methods. The producer invokes these at known transition points in
the message lifecycle. **No aggregation, no sampling, no sliding windows
implemented inside aiokafka** — the user plugs in their preferred backend
(`prometheus_client`, OpenTelemetry, statsd, in-memory test double, etc.).

```python
from typing import Protocol, runtime_checkable

@runtime_checkable
class ProducerMetricsCollector(Protocol):
    """Receive lifecycle events from AIOKafkaProducer.

    All methods are synchronous and called from the hot path; implementations
    should be fast and non-blocking. If asynchronous work is required, defer
    it via an internal queue.

    All time-valued arguments are in **seconds** (consistent with
    ``time.monotonic()`` and the Prometheus naming convention).
    """

    def on_batch_drained(
        self,
        topic: str,
        queue_time_seconds: float,
        batch_size_bytes: int,
        record_count: int,
    ) -> None:
        """Called when a batch leaves the accumulator and is handed to the
        sender. ``queue_time_seconds`` measures time from batch creation
        to drain — equivalent to Java client's ``record-queue-time``."""

    def on_batch_done(
        self,
        topic: str,
        request_latency_seconds: float,
        record_count: int,
    ) -> None:
        """Called when the broker acknowledges the batch. ``request_latency``
        is measured from drain (handoff to sender) to ack — equivalent to
        Java client's ``request-latency``."""

    def on_batch_failure(
        self,
        topic: str,
        exception: BaseException,
        record_count: int,
    ) -> None:
        """Called when a batch ultimately fails (after retries)."""

    def on_buffer_wait(
        self,
        topic: str,
        wait_seconds: float,
    ) -> None:
        """Called when ``add_message`` had to wait for accumulator space —
        equivalent to Java client's ``bufferpool-wait-time``. Direct
        back-pressure signal."""


class NullMetricsCollector:
    """Default no-op implementation. Zero overhead when nothing is plugged in."""
    def on_batch_drained(self, *args, **kwargs): pass
    def on_batch_done(self, *args, **kwargs): pass
    def on_batch_failure(self, *args, **kwargs): pass
    def on_buffer_wait(self, *args, **kwargs): pass
```

Wired into the producer via constructor:

```python
producer = AIOKafkaProducer(
    bootstrap_servers=...,
    metrics_collector=MyCollector(),   # default = NullMetricsCollector()
)
```

### Hook locations

| Callback | File:line | What's measured |
|---|---|---|
| `on_batch_drained` | `message_accumulator.py:288` (`MessageBatch.drain_ready`) | `time.monotonic() - self._ctime` (`_ctime` already exists on line 144) |
| `on_batch_done` | `message_accumulator.py:204` (`MessageBatch.done`) | `time.monotonic() - self._drained_at` (new field set in `drain_ready`) |
| `on_batch_failure` | `message_accumulator.py:248` (`MessageBatch.failure`) | exception type + record count |
| `on_buffer_wait` | `message_accumulator.py:425-429` (`add_message`, around `await batch.wait_drain`) | elapsed time in `wait_drain` |

One new `float` field on `MessageBatch` (`_drained_at`). All other timings
come from data already tracked by the producer.

## Mapping to Java client metrics

Users typically aggregate the raw events into Prometheus `Histogram`s or
similar. The mapping to Java metric names is:

| Java metric | Derived from |
|---|---|
| `record-queue-time-avg/max` | `on_batch_drained.queue_time_seconds` |
| `request-latency-avg/max` | `on_batch_done.request_latency_seconds` |
| `batch-size-avg/max` | `on_batch_drained.batch_size_bytes` |
| `records-per-request-avg` | `on_batch_drained.record_count` |
| `record-send-rate/total` | sum of `on_batch_done.record_count` |
| `record-error-rate/total` | count of `on_batch_failure` |
| `bufferpool-wait-time-ns-total` | sum of `on_buffer_wait.wait_seconds` |

## Why callbacks rather than an embedded Sensor framework

1. **Doesn't reintroduce what #1117 removed.** That PR deleted ~2000 lines
   of dead aggregation code. This proposal adds ~150 lines of callback
   plumbing and no aggregation code.
2. **No new dependencies.** Users who want Prometheus plug in
   `prometheus_client.Histogram`; users who want OTel plug in OTel.
   aiokafka stays dependency-clean.
3. **Better fit for Python ecosystem.** `prometheus_client`, OpenTelemetry
   and statsd all already provide aggregation/percentiles natively — far
   better than the Java `Avg`/`Max` sliding-window stats, which lose
   information.
4. **Unblocks #862 (OpenTelemetry).** With these hooks, OTel
   auto-instrumentation can be written as a third-party package without any
   further changes to aiokafka.
5. **Unblocks #562 (Prometheus exposure).** Documented in the new
   `docs/metrics.rst` with a reference `examples/prometheus_metrics.py`.
6. **Minimal hot-path cost.** With `NullMetricsCollector`, each hook is a
   single attribute lookup + method call returning `None`. ~50 ns in CPython.
   Hooks fire per-batch, not per-record.

## Out of scope for this issue

To keep the PR reviewable, the first PR will include only the
producer-side metrics listed above. Follow-ups can add:
- compression-rate / retry-rate / throttle-time (small additions)
- per-topic dimension via additional labels (`topic` is already passed)
- consumer-side equivalents (`fetch-latency`, lag, commit time)
- OTel-specific helper package (likely belongs in a separate repo)

## Scope of the first PR

- New module `aiokafka/metrics.py` (~80 lines: Protocol + NullCollector + docstrings)
- Constructor parameter `metrics_collector` on `AIOKafkaProducer`
- Hook calls in `MessageBatch.drain_ready`, `MessageBatch.done`,
  `MessageBatch.failure`, `MessageAccumulator.add_message`
- New `_drained_at` field on `MessageBatch`
- `tests/test_metrics.py` with a fake collector verifying:
  - events fire on normal send
  - `on_buffer_wait` fires when accumulator is full
  - `on_batch_failure` fires when delivery fails
  - `NullMetricsCollector` does not introduce overhead (smoke test)
- `docs/metrics.rst` with API reference + Prometheus example
- `examples/prometheus_metrics.py`
- `CHANGES.rst` entry under 0.15.0

Total expected diff: ~300 lines added, ~10 lines modified in existing files.

## Questions for maintainers before I start coding

1. Are you open to this direction in principle?
2. Naming: I went with Java-aligned terminology (`queue_time`, `request_latency`).
   Any preference for renaming?
3. Should `topic` be passed always, or only when an opt-in
   `enable_per_topic_metrics=True` flag is set? Java client emits per-topic
   metrics by default; passing `topic` per-callback adds zero cost but
   downstream aggregation cardinality is the user's choice.
4. Anything about how #1117 was decided that I should be aware of, beyond
   "the old framework was dead code"?

GlebShipilov:
### Changes

Fixes #1166

Adds the callback-based producer metrics collector proposed in #1166.

Summary:
- adds `ProducerMetricsCollector` and `NullMetricsCollector`
- wires batch drain, acknowledgment, failure, and buffer wait callbacks
- documents `metrics_collector` with a Prometheus example

### Checklist

- [x] Add a changelog/news entry
  * Added under `0.15.0` in `CHANGES.rst`; this checkout does not currently have a `CHANGES/` fragment directory.

### Tests

- `python -m ruff format --check aiokafka tests setup.py`
- `python -m ruff check aiokafka tests setup.py`
- `python -m pytest tests/test_metrics.py tests/test_message_accumulator.py`
- `python -m sphinx -E -b html docs docs/_build/html`

### DIFF
--- a/aiokafka/metrics.py
+++ b/aiokafka/metrics.py
@@ -1,8 +1,10 @@
+"""Experimental producer metrics collector API."""
+
 import abc
 
 
 class ProducerMetricsCollector(abc.ABC):
-    """Receive producer batch lifecycle events.
+    """Receive experimental producer batch lifecycle events.
 
     All methods are synchronous and called from the producer hot path.
     Implementations should be fast and non-blocking. If asynchronous work is
@@ -11,10 +13,13 @@ class ProducerMetricsCollector(abc.ABC):
     Exceptions raised by collectors are logged and ignored.
 
     All time-valued arguments are in seconds.
+
+    .. warning::
+        This API is experimental and may change without a deprecation period.
     """
 
     @abc.abstractmethod
-    def on_batch_drained(
+    def on_batch_dispatched(
         self,
         *,
         topic: str,
@@ -22,42 +27,65 @@ class ProducerMetricsCollector(abc.ABC):
         queue_time_seconds: float,
         batch_size_bytes: int,
         record_count: int,
+        attempt: int,
     ) -> None:
-        """Called when a batch leaves the accumulator and is handed to sender."""
+        """Called each time a batch is handed to the sender.
+
+        ``queue_time_seconds`` covers only the queue time for this attempt.
+        ``attempt`` starts at 1. Batch size and record count are reported per
+        attempt and can therefore be observed more than once for retried
+        batches.
+        """
 
     @abc.abstractmethod
-    def on_batch_done(
+    def on_batch_completed(
         self,
         *,
         topic: str,
         partition: int,
-        request_latency_seconds: float,
+        send_to_completion_seconds: float,
         record_count: int,
+        acknowledged: bool,
+        batch_age_seconds: float,
     ) -> None:
-        """Called when the broker acknowledges a batch."""
+        """Called when the producer successfully completes a batch.
+
+        ``acknowledged`` is false when the producer is configured with
+        ``acks=0``. ``batch_age_seconds`` is an upper bound for the latency of
+        records in the batch and includes retries.
+        """
 
     @abc.abstractmethod
-    def on_batch_failure(
+    def on_batch_failed(
         self,
         *,
         topic: str,
         partition: int,
         exception: BaseException,
         record_count: int,
+        batch_age_seconds: float,
     ) -> None:
-        """Called when a batch ultimately fails."""
+        """Called when a batch ultimately fails.
+
+        ``batch_age_seconds`` is an upper bound for the latency of records in
+        the batch and includes retries.
+        """
 
     @abc.abstractmethod
     def on_buffer_wait(
         self, *, topic: str, partition: int, wait_seconds: float
     ) -> None:
-        """Called when appending a record waited for accumulator space."""
+        """Called once when a send operation waited for accumulator space."""
 
 
 class NullProducerMetricsCollector(ProducerMetricsCollector):
-    """Default no-op producer metrics collector."""
+    """Default no-op implementation of the experimental metrics API.
 
-    def on_batch_drained(
+    .. warning::
+        This API is experimental and may change without a deprecation period.
+    """
+
+    def on_batch_dispatched(
         self,
         *,
         topic: str,
@@ -65,26 +93,30 @@ class NullProducerMetricsCollector(ProducerMetricsCollector):
         queue_time_seconds: float,
         batch_size_bytes: int,
         record_count: int,
+        attempt: int,
     ) -> None:
         pass
 
-    def on_batch_done(
+    def on_batch_completed(
         self,
         *,
         topic: str,
         partition: int,
-        request_latency_seconds: float,
+        send_to_completion_seconds: float,
         record_count: int,
+        acknowledged: bool,
+        batch_age_seconds: float,
     ) -> None:
         pass
 
-    def on_batch_failure(
+    def on_batch_failed(
         self,
         *,
         topic: str,
         partition: int,
         exception: BaseException,
         record_count: int,
+        batch_age_seconds: float,
     ) -> None:
         pass
 
--- a/aiokafka/metrics.py
+++ b/aiokafka/metrics.py
@@ -1,5 +1,16 @@
 """Experimental producer metrics collector API."""
 
+import warnings
+
+_CALLBACK_NAMES = frozenset(
+    {
+        "on_batch_completed",
+        "on_batch_dispatched",
+        "on_batch_failed",
+        "on_buffer_wait",
+    }
+)
+
 
 class ProducerMetricsCollector:
     """Receive experimental producer batch lifecycle events.
@@ -11,13 +22,30 @@ class ProducerMetricsCollector:
     Exceptions raised by collectors are logged and ignored.
 
     Each callback has a no-op default implementation. Subclasses only need to
-    override the callbacks they use.
+    override the callbacks they use. Unknown methods starting with ``on_``
+    emit a warning when the subclass is defined to help detect callback name
+    typos.
 
     All time-valued arguments are in seconds.
 
     .. warning::
         This API is experimental and may change without a deprecation period.
     """
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        super().__init_subclass__(**kwargs)
+        unknown_callbacks = sorted(
+            name
+            for name in vars(cls)
+            if name.startswith("on_") and name not in _CALLBACK_NAMES
+        )
+        if unknown_callbacks:
+            warnings.warn(
+                "Unknown producer metrics callbacks (possible typo): "
+                f"{', '.join(unknown_callbacks)}",
+                UserWarning,
+                stacklevel=2,
+            )
 
     def on_batch_dispatched(
         self,
--- a/aiokafka/metrics.py
+++ b/aiokafka/metrics.py
@@ -1,8 +1,7 @@
-from typing import Protocol, runtime_checkable
+import abc
 
 
-@runtime_checkable
-class ProducerMetricsCollector(Protocol):
+class ProducerMetricsCollector(abc.ABC):
     """Receive producer batch lifecycle events.
 
     All methods are synchronous and called from the producer hot path.
@@ -14,45 +13,55 @@ class ProducerMetricsCollector(Protocol):
     All time-valued arguments are in seconds.
     """
 
+    @abc.abstractmethod
     def on_batch_drained(
         self,
+        *,
         topic: str,
+        partition: int,
         queue_time_seconds: float,
         batch_size_bytes: int,
         record_count: int,
     ) -> None:
         """Called when a batch leaves the accumulator and is handed to sender."""
-        ...
 
+    @abc.abstractmethod
     def on_batch_done(
         self,
+        *,
         topic: str,
+        partition: int,
         request_latency_seconds: float,
         record_count: int,
     ) -> None:
         """Called when the broker acknowledges a batch."""
-        ...
 
+    @abc.abstractmethod
     def on_batch_failure(
         self,
+        *,
         topic: str,
+        partition: int,
         exception: BaseException,
         record_count: int,
     ) -> None:
         """Called when a batch ultimately fails."""
-        ...
 
-    def on_buffer_wait(self, topic: str, wait_seconds: float) -> None:
+    @abc.abstractmethod
+    def on_buffer_wait(
+        self, *, topic: str, partition: int, wait_seconds: float
+    ) -> None:
         """Called when appending a record waited for accumulator space."""
-        ...
 
 
-class NullProducerMetricsCollector:
+class NullProducerMetricsCollector(ProducerMetricsCollector):
     """Default no-op producer metrics collector."""
 
     def on_batch_drained(
         self,
+        *,
         topic: str,
+        partition: int,
         queue_time_seconds: float,
         batch_size_bytes: int,
         record_count: int,
@@ -61,7 +70,9 @@ class NullProducerMetricsCollector:
 
     def on_batch_done(
         self,
+        *,
         topic: str,
+        partition: int,
         request_latency_seconds: float,
         record_count: int,
     ) -> None:
@@ -69,11 +80,15 @@ class NullProducerMetricsCollector:
 
     def on_batch_failure(
         self,
+        *,
         topic: str,
+        partition: int,
         exception: BaseException,
         record_count: int,
     ) -> None:
         pass
 
-    def on_buffer_wait(self, topic: str, wait_seconds: float) -> None:
+    def on_buffer_wait(
+        self, *, topic: str, partition: int, wait_seconds: float
+    ) -> None:
         pass
--- a/aiokafka/metrics.py
+++ b/aiokafka/metrics.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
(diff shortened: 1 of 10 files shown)


aiokafka/metrics.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def on_batch_drained(
+        self,
+        topic: str,
+        queue_time_seconds: float,
+        batch_size_bytes: int,
+        record_count: int,
+    ) -> None:
+        """Called when a batch leaves the accumulator and is handed to sender."""
+        ...
```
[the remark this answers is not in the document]
GlebShipilov: cause it's just an interface

aiokafka/producer/message_accumulator.py
```diff
@@ -200,6 +213,9 @@ def done(
         else:
             timestamp_type = 1
 
+        drained_at = self._drained_at
+        should_report = drained_at is not None and not self.future.done()
```
Tayum: Here and further: Why `not` for `self.future.done()`?

aiokafka/metrics.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import annotations
```
Tayum: What do you need it for?

aiokafka/metrics.py
```diff
@@ -0,0 +1,81 @@
...
+    ) -> None:
+        """Called when a batch ultimately fails."""
+        ...
+
+    def on_buffer_wait(self, topic: str, wait_seconds: float) -> None:
+        """Called when appending a record waited for accumulator space."""
+        ...
+
+
+class NullMetricsCollector:
```
Tayum: `NullProducerMetricsCollector`? Or you'd plan to extend it later with other methods from other collectors?

tests/test_metrics.py
```diff
@@ -0,0 +1,327 @@
...
+
+    accumulator.drain_by_nodes(ignore_nodes=[])
+    await add_task
+
+    buffer_wait_events = [
+        event for event in collector.events if event[0] == "buffer_wait"
+    ]
+    assert len(buffer_wait_events) == 1
+    assert buffer_wait_events[0][1] == "test-topic"
+    assert buffer_wait_events[0][2] >= 0
```
Tayum: Is it not:

```suggestion
    assert buffer_wait_events[0][2] > 0
```
?

tests/test_metrics.py
```diff
@@ -0,0 +1,327 @@
...
+
+@pytest.mark.asyncio
+async def test_producer_passes_metrics_collector_to_accumulator():
+    collector = RecordingMetricsCollector()
+    producer = AIOKafkaProducer(metrics_collector=collector)
+    try:
+        assert producer._metrics_collector is collector
+        assert producer._message_accumulator._metrics_collector is collector
+    finally:
+        await producer.stop()
```
Tayum, line 541: It was not started; why would you stop it?

aiokafka/producer/message_accumulator.py
GlebShipilov: self.future is used as the batch's terminal-state marker. We report a terminal metric only while the future is still pending, so repeated or conflicting done()/failure() calls do not emit duplicate metrics. The check is intentionally made before resolving the future.

aiokafka/metrics.py
GlebShipilov: It is not needed in this module, there are no annotations that require deferred evaluation. I'll remove it.

aiokafka/metrics.py
GlebShipilov: I guess, when for example consumer metrics collector will be added, it's better to have 2 different interfaces. I will rename this one

tests/test_metrics.py
GlebShipilov: >= 0 is intentional. The preceding assert not add_task.done() verifies that the coroutine reached the waiting path, while the measured duration may still be 0.0 because of the platform clock resolution. I have already observed this on Windows CI in one of the previous commits, so > 0 would make the test platform-dependent.

tests/test_metrics.py
GlebShipilov: The producer is not started, but it still needs to be closed. Construction sets _closed = False, and leaving it that way triggers the Unclosed AIOKafkaProducer ResourceWarning in __del__. 
stop() supports this case: it skips sender shutdown when no sender task exists and closes the client.

aiokafka/metrics.py
```diff
@@ -0,0 +1,79 @@
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class ProducerMetricsCollector(Protocol):
```
vmaurin: A benefit here to have a Protocol instead of an abstract class ?

I feel it would be more direct and explicit if the implementation contains an explicit import of the aiokafka module + the ProducerMetricsCollector class (unless the protocol exists also in other libs)

aiokafka/metrics.py
```diff
@@ -0,0 +1,79 @@
...
+    required, defer it via an internal queue.
+
+    Exceptions raised by collectors are logged and ignored.
+
+    All time-valued arguments are in seconds.
+    """
+
+    def on_batch_drained(
+        self,
+        topic: str,
```
vmaurin, line 53: maybe the partition could be useful ?

aiokafka/metrics.py
vmaurin: Also trying to see about extensibility, what if tomorrow we want to add an additional parameter to a metric method ?

aiokafka/producer/message_accumulator.py
Tayum: It doesn't allow me to resolve threads... So:

✅ Resolved

tests/test_metrics.py
Tayum: ✅ Resolved

tests/test_metrics.py
Tayum: ✅ Resolved

aiokafka/metrics.py
Tayum: ✅ Resolved

aiokafka/metrics.py
Tayum: ✅ Resolved

aiokafka/metrics.py
GlebShipilov: The benefit of Protocol was structural typing: collector implementations did not have to import or inherit from aiokafka. However, I agree that an explicit extension interface is clearer here and is consistent with interfaces such as ConsumerRebalanceListener. I will change it to an ABC and make NullProducerMetricsCollector its concrete no-op implementation.
                                                                                                                                                             
Neither a Protocol nor an ABC makes adding a required argument backward-compatible. Once released, existing callback signatures should remain stable. Additional metrics can be introduced as new methods with a default no-op implementation, or as part of a major API change.

aiokafka/metrics.py
GlebShipilov: Yes, partition-level metrics would be useful for identifying hot or failing partitions. The accumulator already has the TopicPartition, so I will add partition: int to all callbacks. Collectors can decide whether to use it as a metric label, since doing so may increase cardinality.

aiokafka/metrics.py
vmaurin: Could keyword arguments help a bit ? Like if we do call the provider with `topic=tp.topic,partition=tp.partition` and so on, it could be less order dependent ? But then it is also adding some downside, and maybe some impact on performance ?

I like the idea of a default implementation though :+1:

GlebShipilov: Good suggestion. I’ve updated the callbacks to use keyword-only arguments and changed all call sites accordingly
This removes the dependency on argument order and makes the calls more explicit. The trade-off is that parameter names become part of the public API, but I think that is acceptable for this interface.                                                                                                                                                   

The performance impact should be negligible because these callbacks are invoked per batch, not per record. A local no-op microbenchmark showed a difference of only tens of nanoseconds per call.                                                                                                                                                            
 
NullProducerMetricsCollector remains the default no-op implementation. The existing callbacks are abstract, while future callbacks can be added with a default no-op implementation so existing collector subclasses remain compatible.                                                                                                               

Adding a new argument to an existing callback would still be a breaking change, even with keyword arguments, so those signatures should remain stable after release.

aiokafka/metrics.py
```diff
@@ -0,0 +1,94 @@
...
+    All methods are synchronous and called from the producer hot path.
+    Implementations should be fast and non-blocking. If asynchronous work is
+    required, defer it via an internal queue.
+
+    Exceptions raised by collectors are logged and ignored.
+
+    All time-valued arguments are in seconds.
+    """
+
+    @abc.abstractmethod
```
ods: Does it make sense to make it abstract? As I understand, we may want to implement only relevant to us subset of these methods and "do nothing" is a sensible default behavior for the rest.

GlebShipilov: I guess I can make the methods concrete no-op defaults so users can override only the callbacks they need. This also will make NullProducerMetricsCollector redundant, so the producer can use ProducerMetricsCollector() as its default

aiokafka/metrics.py
```diff
@@ -0,0 +1,103 @@
...
+    .. warning::
+        This API is experimental and may change without a deprecation period.
+    """
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        super().__init_subclass__(**kwargs)
+        unknown_callbacks = sorted(
+            name
+            for name in vars(cls)
+            if name.startswith("on_") and name not in _CALLBACK_NAMES
```
ods, line 40: _nit:_ Probably we can check that the method exists in the base class

ods:
Hi @vmaurin, could you please take a final look? From my point of view, it’s ready to be merged.

vmaurin:
> Hi @vmaurin, could you please take a final look? From my point of view, it’s ready to be merged.

Looks good to me too

PR #1183, opened 2026-09-06
Update & pin github actions in CI

ods:
### Changes

- Update and pin action versions
- Minimal permissions
- `persist-credentials: false`
- Use native caching in setup-python
- Add zizmor to linting

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -309,20 +309,18 @@ jobs:
   test-wheels-aarch64-linux:
     needs: [package-source, package-wheel]
     runs-on: ubuntu-24.04-arm
-    env:
-      py: /opt/python/${{ matrix.pyver }}/bin/python
-      img: quay.io/pypa/manylinux2014_aarch64
 
     strategy:
       matrix:
+        python: ["3.11", "3.12", "3.13", "3.14"]
         include:
-          - pyver: cp311-cp311
+          - python: "3.11"
             aiokafka_whl: dist/aiokafka-*-cp311-cp311-manylinux*_aarch64.whl
-          - pyver: cp312-cp312
+          - python: "3.12"
             aiokafka_whl: dist/aiokafka-*-cp312-cp312-manylinux*_aarch64.whl
-          - pyver: cp313-cp313
+          - python: "3.13"
             aiokafka_whl: dist/aiokafka-*-cp313-cp313-manylinux*_aarch64.whl
-          - pyver: cp314-cp314
+          - python: "3.14"
             aiokafka_whl: dist/aiokafka-*-cp314-cp314-manylinux*_aarch64.whl
 
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -292,44 +309,49 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libkrb5-dev
+      - name: Install python dependencies
+        run: |
+          pip install --upgrade pip setuptools
```
webknjaz: This is likely redundant, at least the setuptools bit.
```suggestion
```

.github/workflows/tests.yml
```diff
@@ -12,38 +12,33 @@ on:
...
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
+      with:
+        persist-credentials: false
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
       with:
         python-version: "3.11"
+        cache: pip  # zizmor: ignore[cache-poisoning]
```
webknjaz, line 31: Personally, I have my own wrapper of the cache action that is Python stability aware before the one in setup-python is not always reliable.

https://github.com/marketplace/actions/cache-python-deps / https://github.com/aio-libs/yarl/commit/6f7c9b032ad1671b619a13ee646f55fc54aef9a0#diff-662dff05c7dfe6681c1007ae601e8573d413a03f9f9d53d951c2cb99caba4fd4R253.

But it's probably fine for now.

.github/workflows/publish.yml
```diff
@@ -292,44 +309,49 @@ jobs:
...
-              yum-config-manager --enable epel && \
-              yum install -y krb5-devel && \
-              pip install --upgrade pip setuptools && \
-              pip install -r requirements-ci.txt && \
-              pip install ${{ matrix.aiokafka_whl }} && \
-              rm -rf aiokafka && \
-              make ci-test-unit && \
-              deactivate'
+          # Remove source code to be sure we use wheel code
+          rm -rf aiokafka
```
webknjaz: Could also set Python in isolated mode..

webknjaz (approved):
Seems fine overall

.github/workflows/publish.yml
ods: Thanks for suggestion! Added isolation. I'm still not sure this way is better, probably I'll revisit it later.

Makefile
```diff
@@ -36,27 +44,28 @@ lint:
 	ruff format --check aiokafka tests setup.py
 	ruff check aiokafka tests setup.py
 	mypy --install-types --non-interactive $(FORMATTED_AREAS)
+	zizmor --pedantic .github/workflows
 
 .PHONY: test
 test: lint
-	pytest -s --show-capture=no --docker-image $(DOCKER_IMAGE) $(FLAGS) tests
+	pytest -s --show-capture=no --docker-image $(DOCKER_IMAGE) tests
```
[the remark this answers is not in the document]
ods: Addressed

